Issue #21, opened 2018-04-09
Refactor modules


tripledes:
Refactoring for:

* Following PEP8 as much as possible
* Make them more modular so easing modifications, tests, ...

tripledes:
Fixed by #24, #26 and #27, closing.

PR #26, opened 2018-04-10
refactor: kubevirt_ovm for python coding style

tripledes:
* 80c line length when possible
* split main() into few less complex methods
* added docstrings

Offline VM module refactored for fixing #21, tested against `tests/ovm.yml`, still to come `kubevirt_vmrs`.

### DIFF
--- a/library/kubevirt_ovm.py
+++ b/library/kubevirt_ovm.py
@@ -22,7 +22,8 @@ author:
 options:
     state:
         description:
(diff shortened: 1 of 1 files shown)


tripledes:
@alexxa thanks for the tips! Should be better now.

alexxa:
@tripledes Yes, looks good, thanks.

PR #44, opened 2018-04-20
issue: made pvc not required

### LINKED ISSUE #41: delete object by name and namespace only
If one wants to delete an object, for example, a VM, only name and namespace should be required, nth else. 
ATM, 
```
    - name: Delete a VM
      kubevirt_vm:
        name: test-vm
        namespace: default
        state: absent

```
returns
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "missing required arguments: pvc"}
```

and the VM is removed only when I specify `pvc`.

tripledes:
* Now when state is absent, pvc is not required, same behaviour as
before hiding registrydisk.
* Also moved all logic, only needed when state is present, after the
right conditional.

Fixes #41

### DIFF
--- a/library/kubevirt_ovm.py
+++ b/library/kubevirt_ovm.py
@@ -60,7 +60,9 @@ options:
         description:
             - "Name of a PersistentVolumeClaim existing in the same namespace
               to use as a base disk for the Offline VM."
-        required: true
+            - "**C(pvc) option is required to create the Offline VM.
+              It's not needed to remove the Offline VM.**"
+        required: false
     src:
         description:
             - "Local YAML file to use as a source to define the Offline VM.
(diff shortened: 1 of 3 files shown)


library/kubevirt_ovm.py
```diff
@@ -60,7 +60,7 @@
         description:
             - "Name of a PersistentVolumeClaim existing in the same namespace
               to use as a base disk for the Offline VM."
-        required: true
+        required: false
```
alexxa: @tripledes I would keep it as `true*`
`* C(pvc) option is required to create the VM. It's not needed to remove the VM.`

And the same for ovm and vmrs.

tripledes: @alexxa it'd be inconsistent with `argument_spec`. If DOC says it's required, it should be required by both states, `present` and `absent`, IMHO.

alexxa: IMHO =) If a user see 'not required' for `present` and doesn't specif it, they will see a msg: "Either pvc or registrydisk is REQUIRED". Ok.. another option, what about `required" false`, and a note in bold: 

`**C(pvc) option is required to create the VM. It's not needed to remove the VM.**`

tripledes:
@alexxa PTAL. I've added the note and also changed the error message so `registrydisk` is not mentioned.

PR #55, opened 2018-06-15
galaxy: prepare modules for Galaxy

tripledes:
* Remove comments and unnecessary fields from metadata file.
* List all modules and their purpose on the contents section.
* Improve instructions, especially to use the modules from Ansible Galaxy.

@alexxa these changes are meant to help usage, especially from Ansible Galaxy, PTAL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,13 @@
 ## Contents
 
 - `lib`: Ansible modules files for KubeVirt management
+    - `kubevirt_raw`: Allow to manage KubeVirt resources, VirtualMachine, OfflineVirtualMachine, VirtualMachineReplicaSet and VirtualMachinePresets.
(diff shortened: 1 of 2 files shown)


alexxa:
@tripledes Sergi, I'm looking at `_facts modules now and think if it's possible to join them, so it will be smth like
```
- kubevirt_facts:
    name: testvmrs
    namespace: vms
    kind: vm/ovm/vmrs
- debug:
    var: kubevirt_{vm/ovm/vmrs}      ---> based on kind
```

alexxa:
s/possible/optimal/efficient ?

tripledes:
@alexxa I wondered that myself and took the current way for being more explicit. Perhaps you could open an issue about that and we'll take this conversion there, what do you think?

PR #59, opened 2018-06-22
make travis to build and publish to Galaxy on tags

tripledes:
* With this change, once a tag is pushed, Travis should run the tests
and notify Ansible Galaxy to import the modules.
* Remove *before_install* as nothing from apt is actually used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ python:
 cache: pip
 env:
   - KUBECONFIG=tests/k8s/config
-before_install:
-  - sudo apt-get update -qq
 install:
   - pip install -r requirements-dev.txt
 script:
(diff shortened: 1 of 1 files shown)


tripledes:
@alexxa just rebased it, PTAL.

.travis.yml
```diff
@@ -15,6 +13,6 @@ script:
   - flake8 --ignore=E402 lib/ansible/module_utils
 branches:
   only:
-    - master
+    - /^v\d+\.\d+(\.\d+)?(-\S*)?$/
```
alexxa: @tripledes please add a comment string why this regex is used

tripledes:
@alexxa added comment with links. PTAL.

PR #62, opened 2018-06-27
facts: unify them

### LINKED ISSUE #56: one or several separate facts modules
ATM, there are four *_facts separate modules for vm, ovm, vmrs, vmpreset (in old naming). Maybe it will be more optimal to join them, so it will be smth like: 

```
- kubevirt_facts:
    name: <name>
    namespace: <namespace>
    kind: vm/ovm/vmrs/vmpreset
- debug:
    var: kubevirt_{vm/ovm/vmrs/vmpreset}      ---> based on kind
```

tripledes:
* Add listing to helper classes.
* Return list of resources instead of single items.
* Fix unit tests.
* Modified playbook examples and add new one.
* Modify README representing the new facts module.

Fixes #56.


tripledes:
@alexxa PTAL. Overall, I think it's a nice improvement. Thanks!

lib/ansible/modules/cloud/kubevirt/kubevirt_facts.py
```diff
@@ -0,0 +1,113 @@
...
+- debug:
+    var: all_vmis
+
+# Gather facts for all VMIRS in vms namespace
+- kubevirt_facts:
+    kind: VirtualMachineInstanceReplicaSet
+    namespace: vms
+  register: all_vms_vmirs
+- debug:
+    var: all_vms_vmirs
```
alexxa: I like this variable naming, it says what kind in what namespace. Notice earlier variable `vm_facts`, its name is kind of general and will be good if a user works with one VM only. Otherwise, I would change it for `testvm_facts` maybe. Just thinking loud, no need to correct it, it's just an example after all =)) 

PS, and in actual tests you used \<name\>_facts.

lib/ansible/modules/cloud/kubevirt/kubevirt_facts.py
```diff
@@ -0,0 +1,113 @@
...
+
+# Gather facts for all VMIS in vms namespace with label 'app = web'
+- kubevirt_facts:
+    kind: VirtualMachineInstance
+    namespace: vms
+    label_selector
```
alexxa: A comment string says `with label 'app = web'`, but I don't see how it works here.

lib/ansible/modules/cloud/kubevirt/kubevirt_facts.py
```diff
@@ -0,0 +1,113 @@
...
+                U(http://www.kubevirt.io/api-reference/master/definitions.html)"
+  returned: On success.
+  type: complex
+  contains:
+    api_version:
+      description: The versioned schema of this representation of an object.
+      returned: success
+      type: str
+    kind:
+      description: Represents the REST resource this object represents.
```
alexxa: Too many `represents` in one sentence. Maybe `The REST resource of the object.`?

lib/ansible/modules/cloud/kubevirt/kubevirt_facts.py
```diff
@@ -0,0 +1,113 @@
...
+      returned: success
+      type: str
+    metadata:
+      description: "Standard object metadata. Includes name, namespace,
```
alexxa: Please no sentences without subject: `Standard object metadata including name, ....`

lib/ansible/modules/cloud/kubevirt/kubevirt_facts.py
```diff
@@ -0,0 +1,113 @@
...
+items:
+  description: "Dictionary describing the requested resource.
+                Resource attributes are mapped to dictionary
+                keys, which can be found at the following URL:
+                U(http://www.kubevirt.io/api-reference/master/definitions.html)"
+  returned: On success.
+  type: complex
+  contains:
+    api_version:
+      description: The versioned schema of this representation of an object.
```
alexxa: The versioned schema of the object representation.

lib/ansible/modules/cloud/kubevirt/kubevirt_facts.py
```diff
@@ -0,0 +1,113 @@
...
+      description: Represents the REST resource this object represents.
+      returned: success
+      type: str
+    metadata:
+      description: "Standard object metadata. Includes name, namespace,
+                    annotations, labels, etc."
+      returned: success
+      type: dict
+    spec:
+      description: "Specific attributes of the object. Will vary based
```
alexxa: `Will` doesn't have a subject. `Specific attributes of the object depending on the I(api_version) and I(kind).`

tripledes:
@alexxa made the modifications, thanks for reviewing. :+1:

Issue #61, opened 2018-06-26
Move modules using patch() to KubeVirt Python client


tripledes:
Now that https://github.com/kubevirt/client-python/issues/20 seems to be fixed, `kubevirt_scale_vmirs` and `kubevirt_vm_status` should use KubeVirt's Python client and avoid Kubernetes when possible.

tripledes:
Blocked by https://github.com/kubevirt/client-python/issues/24

vatsalparekh:
@mmazur @machacekondra @pkliczewski now that we moved to dynamic client, is this issue valid anymore?

machacekondra:
It's not. IMHO can be closed.

PR #68, opened 2018-07-05
[WIP] make the specialized modules to use kubevirt python client

tripledes:
Initial work for #61


tripledes:
Closing until it can be properly fixed.

PR #80, opened 2018-08-25
Support Persistent Volume Claim actions

masayag:
The PR adds support for PVC actions such as create, remove and receive
facts of the PVC.

This PR requires the k8s endpoint in order to perform PVC actions using
hte PVC resources.

Also, it introduces the ability to wait till the pvc creation is
completed.

### DIFF
--- a/lib/ansible/module_utils/k8svirt/common.py
+++ b/lib/ansible/module_utils/k8svirt/common.py
@@ -9,6 +9,9 @@ import os
 import sys
 import yaml
 import kubevirt as sdk
+from kubernetes import config, client as core_client
+from kubernetes.config import kube_config, ConfigException
+from kubernetes.client import Configuration
 
 from ansible.module_utils.basic import AnsibleModule
 from ansible.module_utils.six import iteritems
@@ -17,8 +20,6 @@ if hasattr(sys, '_called_from_test'):
     from helper import AUTH_ARG_SPEC
 else:
     from ansible.module_utils.k8svirt.helper import AUTH_ARG_SPEC
-
-from kubernetes.config import kube_config, ConfigException
 
 
 class K8sVirtAnsibleModule(AnsibleModule):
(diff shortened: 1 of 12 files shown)


lib/ansible/module_utils/k8svirt/helper.py
```diff
@@ -379,3 +395,81 @@ def replace(self, body, namespace, name):
...
+            namespace=namespace, body=pvc_body)
+        if not wait:
+            return created_pvc
+
+        name = pvc_body['metadata']['name']
+        start = time.time()
+        while time.time() < start + timeout:
+            entity = self.__client.read_namespaced_persistent_volume_claim(
+                name, namespace)
+            if entity.status.phase == 'Succeeded':
```
masayag: @pkliczewski what is the expected status in this context ? According to https://github.com/ansibleplaybookbundle/import-vm-apb/blob/master/roles/import-from-url/tasks/provision.yml#L66 it should be Bound.

lib/ansible/module_utils/k8svirt/common.py
```diff
@@ -72,32 +73,60 @@ def __configure_by_file(self):
...
+            client_configuration=sdk.configuration
+        )
+        return sdk.DefaultApi()
+
     def __configure_by_params(self):
         """ Return API client from configuration file """
         auth_args = AUTH_ARG_SPEC.keys()
+        core_configuration = Configuration()
 
         for key, value in iteritems(self.params):
```
pkliczewski: Would be good to extract is as a method which would accept configuration as param.

lib/ansible/module_utils/k8svirt/helper.py
```diff
@@ -379,3 +395,81 @@ def replace(self, body, namespace, name):
...
+        pvc_body = core_sdk.V1PersistentVolumeClaim().to_dict()
+        pvc_body.update(copy.deepcopy(body))
+        created_pvc = self.__client.create_namespaced_persistent_volume_claim(
+            namespace=namespace, body=pvc_body)
+        if not wait:
+            return created_pvc
+
+        name = pvc_body['metadata']['name']
+        start = time.time()
+        while time.time() < start + timeout:
```
pkliczewski: Why not to use a watch here? In similar way we do it [here](https://github.com/kubevirt/client-python/blob/master/examples/example_vms.py#L34)

lib/ansible/module_utils/k8svirt/helper.py
pkliczewski: Based on the logic from `import-vm-apb` we need to wait on status `Bound` and later on import_status to be either 'Succeeded' or 'Failed'. This makes it a bit more complicated because first we wait on creation and later on import. That is why I suggested to have a separate waiting task like in `import-vm-apb`.

lib/ansible/modules/cloud/kubevirt/kubevirt_raw.py
```diff
@@ -195,6 +195,21 @@
...
+           returned: success
+           type: bool
+        timeout:
+           description: Specifies how much time in seconds to wait for an
+                        action to compelte if 'wait' option is enabled.
+           returned: success
+           type: int
+        interval:
+           description: Specifies how the interval in seconds between polling
+                        of the watched entity
```
alexxa: "Specifies the length of the interval in seconds between polling the watched entity" or maybe "Specifies the frequency of polling the watched entity in seconds"

tests/playbooks/kubevirt_raw_pvc_remove.yml
```diff
@@ -0,0 +1,17 @@
+---
+- name: Removes a persistent volume claim
```
alexxa: Remove

masayag: fixed.

lib/ansible/module_utils/k8svirt/common.py
pkliczewski: We can improve this later

PR #91, opened 2018-09-12
Update playbook description

masayag:


masayag:
The example in this PR is duplication of tests/playbooks/kubevirt_pvcs_facts.yml differ only by the namespace. Therefore it should be removed.

PR #92, opened 2018-09-12
Remove duplicated example

masayag:
The removed example is a duplication of
tests/playbooks/kubevirt_pvcs_facts.yml

It is differ only by the namespace value. Therefore it should be removed.


masayag:
@vatsalparekh even with changing the namespace, this yml is the same as https://github.com/kubevirt/ansible-kubevirt-modules/blob/master/tests/playbooks/kubevirt_pvcs_facts.yml except of the namespace value. therefore we can remove this one.

PR #96, opened 2018-09-13
[RFR] changes in raw playbooks

vatsalparekh:


tests/playbooks/kubevirt_raw_pvc.yml
```diff
@@ -27,3 +28,13 @@
            resources:
              requests:
                storage: 20Mi
+    - name: Delete pvc-demo PVC
```
masayag: I'm not sure about adding the deletion part to this one - as an example, the user won't be able to evaluate what was created, since it will be removed instantly.

Also, if the purpose is to create and remove, we don't really need to wait for the completion of the previous step.

vatsalparekh: @masayag agree
But as this example is also a test, and we would want to take these test playbooks to run in CI, it makes sense to be testing deletion as well
About wait, we need to test that it works, so I think it makes sense to keep it there
your thoughts on this @pkliczewski ?

masayag: I'd separate the two things: tests designed for CI should be located on a dedicated folder (or repository) and examples for the user in a different location.
Each test has a different goal - one serves as an example for users to play with and getting familiar with capabilities of the module and the automation or CI tests should be more comprehensive and cover the various functionalities of the module.

IIUC, the examples located under the tests folder is designed to serve as examples for the users.

vatsalparekh: Sounds good, changing it to a separate file

PR #98, opened 2018-09-14
Use k8s modules helpers using dynamic client

machacekondra:
- Removed few unrelevant tests from `module_utils`, as it's now using k8s `module_utils`, but maybe we still should add some relevant tests for our cases.
 - Requires Ansible 2.7 & openshift client 0.6.2

Signed-off-by: Ondra Machacek <omachace@redhat.com>


lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
```diff
@@ -193,56 +191,146 @@ def execute_module(self):
...
+        stream = None
+        try:
+            w = watch.Watch()
+            w._api_client = self.client.client
+            if namespace:
+                stream = w.stream(resource.get, serialize=False, namespace=namespace, timeout_seconds=wait_time)
+            else:
+                stream = w.stream(resource.get, serialize=False, namespace=namespace, timeout_seconds=wait_time)
+        except KubernetesException:
+            pass
```
pkliczewski: We should not pass here. We should fail with information that we failed to create watch.

lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
```diff
@@ -193,56 +191,146 @@ def execute_module(self):
...
-        if not self.params.get('verify_ssl'):
-            setattr(config, 'verify_ssl', False)
-
-        return kubernetes.client.ApiClient(configuration=config)
+
+        if wait and stream is not None:
+            return_obj = self._read_stream(resource, w, stream, name, replicas)
+
+        if not return_obj:
+            return_obj = self._wait_for_response(name, namespace)
```
pkliczewski: Why do we need to wait if we can use events to understand whether the cluster got to required state?

lib/ansible/modules/cloud/kubevirt/kubevirt_vm_status.py
```diff
@@ -132,127 +132,94 @@
...
-            self.exit_json(changed=True, result=api_response)
-        except ApiException as exc:
-            self.fail_json(msg='Failed to manage requested object',
-                           error=exc.reason)
+        state = self.params.get('state')
+        resource_version = self.params.get('resource_version')
+        name = definition['metadata']['name']
+        namespace = definition['metadata'].get('namespace')
+        api_version = self.params.get('apiVersion', 'kubevirt.io/v1alpha2')
+        kind = self.params.get('kind', 'VirtualMachine')
```
pkliczewski: Why do we need this. We know what is the kind for this piece of code.

lib/ansible/modules/cloud/kubevirt/kubevirt_vm_status.py
```diff
@@ -132,127 +132,94 @@
...
+
+        if hasattr(existing.spec, 'running'):
+            existing_running = existing.spec.running
+
+        if resource_version and resource_version != existing.metadata.resourceVersion:
+            self.exit_json(**return_attributes)
+
+        # TODO: Not nice code, but working, just not to break, what have been done,
+        # but really need rewrite.
+        if state == 'running':
```
pkliczewski: What about other states like `Pending` or `Scheduling`?

lib/ansible/modules/cloud/kubevirt/kubevirt_vm_status.py
```diff
@@ -132,127 +132,94 @@
...
+            if existing_running:
```
pkliczewski: Lines 200-209 and 211-222 are very similar with very minor changes. Looks like good candidate to extract a method.

masayag:
@machacekondra Could you include couple of yml files (create/collect facts) to see how the new API should look like and test it locally ?

lib/ansible/modules/cloud/kubevirt/kubevirt_vm_status.py
machacekondra: That's something we need to do. Currently the modules has same functionality as before. So we don't break it. The improvement patch can be done later.

lib/ansible/modules/cloud/kubevirt/kubevirt_vm_status.py
machacekondra: Yes, right.

machacekondra:
> @machacekondra Could you include couple of yml files (create/collect facts) to see how the new API should look like and test it locally ?

It's same as before, I didn't want to break backward compatibility for now.

lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
machacekondra: Ok

lib/ansible/modules/cloud/kubevirt/kubevirt_vm_status.py
machacekondra: Right, I guess we don't have to provide user the option to override it.

lib/ansible/modules/cloud/kubevirt/kubevirt_vm_status.py
pkliczewski: OK

lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
```diff
@@ -193,56 +191,146 @@ def execute_module(self):
...
+        """ Create a stream of events for the object """
+        w = None
+        stream = None
+        try:
+            w = watch.Watch()
+            w._api_client = self.client.client
+            if namespace:
+                stream = w.stream(resource.get, serialize=False, namespace=namespace, timeout_seconds=wait_time)
+            else:
+                stream = w.stream(resource.get, serialize=False, namespace=namespace, timeout_seconds=wait_time)
```
masayag: The two calls are the same (lines 285 and 286) with or without namespace provided, either we don't need the if-else or in case of namespace not being provided, the call should be different.

lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
machacekondra: In case the `wait=false`

pkliczewski: My point was that we use `time.sleep(2)` and we poll. I thought it would be better to rely our logic on   a `watch` instead.

machacekondra:
@pkliczewski @masayag Ready to review.

lib/ansible/modules/cloud/kubevirt/kubevirt_raw.py
```diff
@@ -20,99 +20,37 @@
...
+
 options:
+  merge_type:
+    description:
+    - Whether to override the default patch merge approach with a specific type. By the default, the strategic
+      merge will typically be used.
+    - For example, Custom Resource Definitions typically aren't updatable by the usual strategic merge. You may
+      want to use C(merge) if you see "strategic merge patch format is not supported"
+    - See U(https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/#use-a-json-merge-patch-to-update-a-deployment)
+    - Requires openshift >= 0.6.2
```
pkliczewski: I think we need to make it more strict since KubeVirt was not there. Let's do it in separate PR.

lib/ansible/modules/cloud/kubevirt/kubevirt_raw.py
```diff
@@ -20,99 +20,37 @@
...
-            - Path to a certificate used to authenticate with the API.
-    key_file:
-        description:
-            - Path to a key file used to authenticate with the API.
 
 requirements:
-    - python >= 2.7
-    - kubevirt-python >= 0.4.1-132-g074b4658
+  - python >= 2.7
+  - openshift >= 0.6.2
```
pkliczewski: As above in separate PR

lib/ansible/modules/cloud/kubevirt/kubevirt_raw.py
```diff
@@ -207,12 +145,98 @@
...
+            self.exit_json(**results[0])
+
+        self.exit_json(**{
+            'changed': changed,
+            'result': {
+                'results': results
+            }
+        })
+
+    # TODO: Candidate to be common method:
```
pkliczewski: Let's do it in separate PR

lib/ansible/modules/cloud/kubevirt/kubevirt_raw.py
```diff
@@ -207,12 +145,98 @@
...
+        """ Wait for ready_replicas to equal the requested number of replicas. """
+        return_obj = None
+        for event in stream:
+            if event.get('object'):
+                entity = ResourceInstance(resource, event['object'])
+                metadata = entity.metadata
+                if entity.status.phase == 'Bound':
+                    annotations = metadata.annotations if \
+                        metadata.annotations else {}
+                    IMPORT_STATUS_KEY = 'cdi.kubevirt.io/storage.import.pod.phase'
```
pkliczewski: Would be good to extract it as constant

lib/ansible/modules/cloud/kubevirt/kubevirt_raw.py
```diff
@@ -207,12 +145,98 @@
...
+                    if (not self._use_cdi(annotations, labels) or
+                            import_status == 'Succeeded'):
+                        w.stop()
+                        return entity
+                    elif entity.status.phase == 'Failed':
+                        w.stop()
+                        self.fail_json(msg="Failed to import PersistentVolumeClaim")
+
+    def _use_cdi(self, annotations, labels):
+        IMPORT_ENDPOINT_KEY = 'cdi.kubevirt.io/storage.import.endpoint'
```
pkliczewski: Would be good to extract it as constant

lib/ansible/modules/cloud/kubevirt/kubevirt_raw.py
```diff
@@ -207,12 +145,98 @@
...
+        endpoint = annotations.get(IMPORT_ENDPOINT_KEY)
+        app_label = labels.get('app')
+        return endpoint and app_label == 'containerized-data-importer'
```
pkliczewski: Would be good to extract it as constant

lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
```diff
@@ -15,12 +15,12 @@
...
 short_description: Scale up or down KubeVirt VMI ReplicaSet
 
-version_added: "2.5"
+version_added: "2.8"
 
 author: KubeVirt Team (@kubevirt)
 
 description:
-  - "Use Kubernets Python SDK to scale up or down KubeVirt
+  - "Use Openshift Python SDK to scale up or down KubeVirt
```
pkliczewski: Kubevirt can be deployed in both Kubernetes and Openshift cluster. Let's keep both here.

lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
```diff
@@ -39,57 +39,13 @@
...
-            - Path to a key file used to authenticate with the API.
 
+extends_documentation_fragment:
+  - k8s_auth_options
 
 requirements:
-    - python >= 2.7
-    - kubernetes python client >= 6.0.0
+  - python >= 2.7
+  - openshift >= 0.6.2
```
pkliczewski: As above

lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
```diff
@@ -132,117 +88,169 @@
...
-
-        if not self.params.get('verify_ssl'):
-            setattr(config, 'verify_ssl', False)
-
-        return kubernetes.client.ApiClient(configuration=config)
+
+        if wait and stream is not None:
+            return_obj = self._read_stream(resource, w, stream, name, replicas)
+
+        if not return_obj:
```
pkliczewski: I still do not understand this if. Let's say that `wait` is `None` and we won't read stream. As a result `return_obj` will be None so we will ?? wait. My question is why should we if said we do not want to by not setting `wait`.

lib/ansible/modules/cloud/kubevirt/kubevirt_vm_status.py
```diff
@@ -132,127 +89,88 @@
...
-                group, version, namespace, plural, name, body)
-            self.exit_json(changed=True, result=api_response)
-        except ApiException as exc:
-            self.fail_json(msg='Failed to manage requested object',
-                           error=exc.reason)
+        state = self.params.get('state')
+        resource_version = self.params.get('resource_version')
+        name = definition['metadata']['name']
+        namespace = definition['metadata'].get('namespace')
+        api_version = self.params.get('apiVersion', 'kubevirt.io/v1alpha2')
```
pkliczewski: Let's use constant for it.

requirements-dev.txt
```diff
@@ -1,6 +1,6 @@
 pytest >= 3.6.0, < 4.0.0
 mock >= 2.0.0, < 3.0.0
 flake8 >= 3.5.0, < 4.0.0
-ansible>=2.4.0, < 3.0.0
-git+https://github.com/kubevirt/client-python.git@v0.1.0
-kubernetes >= 6.0.0, < 7.0.0
+# Fixme, when Ansible 2.7 is released: ansible >= 2.7.0, < 3.0.0
+git+https://github.com/ansible/ansible.git@v2.7.0rc3
+openshift >= 0.6.2, < 1.0.0
```
pkliczewski: is it openshift client or openshift cluster? I am confused about the version.

machacekondra: client

pkliczewski: Oh ok, can we add a comment?

lib/ansible/modules/cloud/kubevirt/kubevirt_scale_vmirs.py
machacekondra: I've removed it, we now return the object returned by `patch_resource` method.

lib/ansible/modules/cloud/kubevirt/kubevirt_raw.py
```diff
@@ -24,95 +24,34 @@
...
-        description:
-            - Path to a certificate used to authenticate with the API.
-    key_file:
-        description:
-            - Path to a key file used to authenticate with the API.
 
 requirements:
-    - python >= 2.7
-    - kubevirt-python >= 0.4.1-132-g074b4658
+  - python >= 2.7
```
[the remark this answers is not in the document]
machacekondra: AFAIK in requirments there are only pip requirments, not system requirments.

requirements-dev.txt
machacekondra: In general in Ansible modules add dependencies on SDKs/client libs. the SDKs client libs need to achieve compatibility with server, so it's common it's requierment to SDK/client.

pkliczewski: OK

machacekondra:
@pkliczewski Should I solve some other stuff, as part of this PR, or should another stuff go to another PR?

pkliczewski:
@machacekondra The change looks good but travis failed. Once it will be green again I will merge.

machacekondra:
> @machacekondra The change looks good but travis failed. Once it will be green again I will merge.

@pkliczewski Done.

pkliczewski:
@machacekondra Thank you! :tada:

PR #104, opened 2018-09-27
Move everything into directories aligned with upstream

mmazur:
Previous directory structure would conflict with upstream ansible. This one doesn't.
Note: the modules are now ansible.modules.clustering.kubevirt, not .cloud.kubevirt


mmazur:
Turns out conftest.py can be dropped.

PR #97, opened 2018-09-14
Change directory structure to be a proper role

machacekondra:
This PR is changing the structure of the repo to be a proper role:

  - It moves the helper modules to `module_utils` directory, so user don't need to set those as env `ANSIBLE_MODULE_UTILS` variable.
 - It moves the modules to `library` directory, so user don't need to set those as `ANSIBLE_LIBRARY` variable.

Now it's OK if user just use the role before any other use of the modules, without setting the env vars.

For example:

```yaml
---
- name: Load modules
  hosts: localhost

  roles:
    - role: kubevirt.kubevirt-modules
      install_python_requirements: no

- hosts: localhost

  tasks:
    - name: Fetch VM 'myvm'
      kubevirt_facts:
        name: myvm
        namespace: vms
        kind: VirtualMachine
```

Signed-off-by: Ondra Machacek <omachace@redhat.com>

PR #105, opened 2018-09-27
Move everything into directories aligned with upstream

mmazur:
Previous directory structure would conflict with upstream ansible. This one doesn't.
Note1: the modules are now ansible.modules.clustering.kubevirt, not .cloud.kubevirt
Note2: conftest.py is no longer needed.


pkliczewski:
@machacekondra please review

mmazur:
@machacekondra please ignore the others, this is the final form.

machacekondra:
@pkliczewski @mmazur If we are going to merge https://github.com/kubevirt/ansible-kubevirt-modules/pull/97 I think we don't need this PR. So it's up to you to decide whether we want to have a structure as a role or a structure which reflect Ansible git.

mmazur:
If the long–term goal is to merge with upstream, then we just have to deal with any issues that arise from having the upstream directory structure. For comparison with upstream ansible:
- https://github.com/ansible/ansible/tree/stable-2.7/lib/ansible/modules/clustering – where _k8s_ and _openshift_ modules reside; and where _kubevirt_ should go
- https://github.com/ansible/ansible/tree/stable-2.7/lib/ansible/modules/cloud/openstack – where _openstack_ module resides
- https://github.com/ansible/ansible/tree/stable-2.7/test/units/modules/cloud/openstack – and tests for the _openstack_ module

The earlier our code reflects the final directory structure, the better, as to not to cause confusion in any possible downstream code.

Heh, just noticed that the directory name is _test/_, not __tests/__, but I'll fix that in a separate PR.

pkliczewski:
Thanks! :tada:

machacekondra:
Long-term is actually having modules in galaxy, that's where the galaxy is currently aiming. So the long-term structure is unknown. But I think it doesn't really matter, from user perspective I think it's better to have the role structure of our repo, because it's easier to setup to use it. From devel perspective it doesn't matter. As we don't have any users so far, it's ok to use this structure I think.

PR #106, opened 2018-10-03
Actually utilize KubeVirtVM class

mmazur:
The module was using the upstream k8s class by accident.


lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
```diff
@@ -248,10 +248,5 @@ def _use_cdi(self, annotations, labels):
         return endpoint and app_label == 'containerized-data-importer'
 
 
-def main():
```
pkliczewski: It seems like keeping main method as entry point is a common [practice](https://github.com/kubernetes-incubator/kubespray/blob/master/library/kube.py#L313). Please only change line 253

mmazur:
Ok, corrected.

PR #107, opened 2018-10-05
Re-add KubeVirtVM() tests with dynamic client

mmazur:
I've finally figured out how to get the new dyn client tested properly. As usual the end result looks somewhat trivial and does not at all indicate the amount of head scratching and dead ends that went into its creation. But that's likely a good sign – if the solution seems simple, that's a good indication that it's the right one.

Please don't merge this PR, it's just for getting early comments. Come Monday I'll add the missing tests (and there were a few in the pre–dyn client code) and likely generalize the code a bit while I'm at it. Then I'll send the final PR with all the tests for this class.

### DIFF
--- a/tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
+++ b/tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
@@ -1,0 +1,94 @@
+import sys
+import pytest
+
+from ansible.compat.tests.mock import patch, MagicMock
+
+from ansible.module_utils.k8s.common import K8sAnsibleMixin
+from openshift.dynamic import Resource
+
+from utils import set_module_args, AnsibleExitJson, exit_json
+
+# FIXME: paths/imports should be fixed before submitting a PR to Ansible
+sys.path.append('lib/ansible/modules/clustering/kubevirt')
+
+import kubevirt_raw as mymodule
+
+RESOURCE_DEFAULT_ARGS = { 'api_version': 'v1', 'group': 'kubevirt.io',
+                            'prefix': 'apis', 'namespaced': True }
+TESTABLE_KINDS = ( 'VirtualMachineInstance', 'VirtualMachine', 'VirtualMachineInstanceReplicaSet',
+                    'VirtualMachineInstancePreset', 'PersistentVolumeClaim' )
+ 
+
+class TestKubeVirtRawModule(object):
+    @pytest.fixture(autouse=True)
+    def setup_class(cls, monkeypatch):
+        monkeypatch.setattr(
+            mymodule.KubeVirtVM, "exit_json", exit_json)
+        # Create mock methods in Resource directly, otherwise dyn client
+        # tries binding those to corresponding methods in DynamicClient
+        # (with partial()), which is more problematic to intercept
+        Resource.get = MagicMock()
+        Resource.create = MagicMock()
+        Resource.delete = MagicMock()
+        # Globally mock some methods, since all tests will use this
+        K8sAnsibleMixin.get_api_client = MagicMock()
+        K8sAnsibleMixin.get_api_client.return_value = None
+        K8sAnsibleMixin.find_resource = MagicMock()
+
+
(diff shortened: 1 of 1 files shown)


tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
```diff
@@ -0,0 +1,66 @@
...
+    pass
+
+
+def exit_json(*args, **kwargs):
+    if 'changed' not in kwargs:
+        kwargs['changed'] = False
+    raise AnsibleExitJson(kwargs)
+
+
+class TestKubeVirtRawModule(object):
```
pkliczewski: I think that the class and two methods could be moved to utility module which could be use across tests.

tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
```diff
@@ -0,0 +1,66 @@
...
+
+import kubevirt_raw as mymodule
+
+
+def set_module_args(args):
+    args = json.dumps({'ANSIBLE_MODULE_ARGS': args})
+    basic._ANSIBLE_ARGS = to_bytes(args)
+
+
+class AnsibleExitJson(Exception):
```
pkliczewski: Why not to use `exceptions.NotFoundError` from the dynamic client.

tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
mmazur: The glue code is copy&pasted from the two other existing test files. I'd prefer to finish this code as is, get it merged, then have a separate PR for extracting any commonalities across all the test code.

tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
mmazur: Dunno. I'll make note to take a look into it after I've written everything.

tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
pkliczewski: We should not copy the code. This is perfect place to push a new PR which would make the code common and we can reuse it here.

tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
pkliczewski: It is much better to look for errors which can actually occur in the code.

tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
mmazur: Ok, that order it will be then.

tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
mmazur: I've looked into it. That exception is not an error, it's a method of injecting into the underlying AnsibleModule base class an exit point for our code to take back control. It's the correct option to explicitly define an arbitrary exception in this case – using anything else would falsely imply there is some other meaning behind it other than "we need the underlying module to stop executing and give us back control at this precise point". Which there isn't.

pkliczewski: OK

tests/units/modules/clustering/kubevirt/test_kubevirt_raw.py
```diff
@@ -0,0 +1,66 @@
...
+
+from openshift.dynamic import Resource
+
+# FIXME: paths/imports should be fixed before submitting a PR to Ansible
+sys.path.append('lib/ansible/modules/clustering/kubevirt')
+
+import kubevirt_raw as mymodule
+
+
+def set_module_args(args):
```
pkliczewski: Please update the PR and use the utils module and remove both methods and exception.

mmazur:
So, this is how it's going to look like. A few tests left and I'll be done.

pkliczewski:
@machacekondra please review

PR #109, opened 2018-10-11
Unify exception handling across modules

mmazur:
Make sure all the _read_stream() methods have the same basic structure.

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
@@ -148,6 +148,7 @@ result:
 '''
 
 import copy
+import traceback
 
 from ansible.module_utils.k8s.common import COMMON_ARG_SPEC
 from ansible.module_utils.k8s.raw import KubernetesRawModule
@@ -249,8 +250,11 @@ class KubeVirtVM(KubernetesRawModule):
 
 
 def main():
-    '''Entry point.'''
-    KubeVirtVM().execute_module()
+    module = KubeVirtVM()
+    try:
+        module.execute_module()
+    except Exception as e:
+        module.fail_json(msg=str(e), exception=traceback.format_exc())
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
```diff
@@ -223,23 +223,26 @@ def _create_stream(self, resource, namespace, wait_time):
...
+                        import_status = annotations.get(IMPORT_STATUS_KEY)
+                        labels = metadata.labels if metadata.labels else {}
+                        if (not self._use_cdi(annotations, labels) or
+                                import_status == 'Succeeded'):
+                            watcher.stop()
+                            return entity
+                        elif entity.status.phase == 'Failed':
+                            watcher.stop()
+                            self.fail_json(msg="Failed to import PersistentVolumeClaim")
+        except Exception as exc:
```
pkliczewski: What kind of exceptions we would get here?

mmazur: No clue. But the other class does this inside its _read_stream() and it looks like a good idea in general, since if we were to have any kinds of exceptions, this would create a mildly more informative error message. And improving error reporting is one of the open issues.

pkliczewski: Which class? Can we reuse?

mmazur: https://github.com/kubevirt/ansible-kubevirt-modules/blob/master/lib/ansible/modules/clustering/kubevirt/kubevirt_scale_vmirs.py#L211 and no, we can't. From what I've looked at the code, it does not make sense to try to separate stream handling into a helper class. It's better to just make sure the stream handling code in both classes behave similarly (hence this patch) and then write some universal testing code to test both classes with. (I'm referring to the Watch thing.)

machacekondra: What I did in ovirt modules, and found quite good, is to have just one try, except per module and re-reaise with traceback, that way debugging is OK, and code is not verbose.

pkliczewski: @machacekondra Please point Mariusz to that code so we can do the same here

mmazur: Yup, that looks better. Just pushed a change making all the modules have exception handling that way.

PR #110, opened 2018-10-12
Use 'wait_timeout' instead of 'wait' arg

mmazur:
Commit 1: Bring back 'timeout' variable instead of 'wait_timeout'

Two options:
1. Switch to the new name – 'wait_timeout' – but then I need to modify existing playbooks to use the new var name.
2. Bring back the old var name, which is what this commit does.

I assume @machacekondra had a reason to switch to wait_timeout and it does sound cleaner. So, what do we do?

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
@@ -140,9 +140,9 @@ result:
                         till it completes.
            returned: success
            type: bool
-        timeout:
+        wait_timeout:
            description: Specifies how much time in seconds to wait for an
-                        action to compelte if 'wait' option is enabled.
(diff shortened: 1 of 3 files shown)


pkliczewski:
Let's make pvc great again :)

machacekondra:
Wait timeout was used because k8s does use it, so I wanted to be consistent.

mmazur:
That's a good reason, let's switch to that.

mmazur:
Ready for merge. This plus PR 112 make PVC work correctly at least to the (narrow) extent that I've tested it.

mmazur:
Wait, no, forgot something.

mmazur:
Yeah, please merge this as–is (along with all the other PRs). I'll get a fix for the Waiting logic for PVCs in a separate PR, since that'll require fixes in the tests as well.

PR #112, opened 2018-10-12
Handle 'merge_type' var as required

mmazur:
For KubeVirtVM() to handle updating existing objects, merge_type must be properly handled. It wasn't before.

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
@@ -162,7 +162,8 @@ except ImportError as exc:
 
 RAW_ARG_SPEC = {
     'wait': {'type': 'bool', 'default': True},
-    'wait_timeout': {'type': 'int', 'default': 20}
+    'wait_timeout': {'type': 'int', 'default': 20},
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']},
 }
 
 


mmazur:
Yeah, please merge this as–is. I'll get a fix for the Waiting logic for PVCs in a separate PR, since that'll require fixes in the tests as well.

lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
```diff
@@ -174,6 +174,7 @@ def __init__(self, *args, **kwargs):
     def argspec(self):
         argspec = copy.deepcopy(COMMON_ARG_SPEC)
         argspec.update(copy.deepcopy(RAW_ARG_SPEC))
+        argspec['merge_type'] = dict(type='list', choices=['json', 'merge', 'strategic-merge'])
```
machacekondra: I would add it to RAW_ARG_SPEC

mmazur: Done.

lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
```diff
@@ -162,7 +162,8 @@ class KubernetesException(Exception):
 
 RAW_ARG_SPEC = {
     'wait': {'type': 'bool', 'default': True},
-    'wait_timeout': {'type': 'int', 'default': 20}
+    'wait_timeout': {'type': 'int', 'default': 20},
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']}
```
machacekondra: Not sure what python convention we should use, but I like to have ',' at the end, so the path is more easy to read next time we add new parameter.

machacekondra: And also, what about adding a default?

mmazur:
Can't find your comments in the UI, so I answer here:
1. Added a trailing comma to the RAW_ARG_SPEC definition.
2. Default _merge_type_ is handled internally by KubernetesRawModule, it's not defined statically in the argument definition ([relevant code](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/k8s/raw.py#L42)). I believe the idea is to let the module have sane behavior depending on changing conditions and not have to commit to any particular way of doing things. Quoting from the docs: _Whether to override the default patch merge approach with a specific type. By the default, the strategic merge will **typically** be used._

machacekondra:
> Can't find your comments in the UI, so I answer here:

    Added a trailing comma to the RAW_ARG_SPEC definition.
    Default merge_type is handled internally by KubernetesRawModule, it's not defined statically in the argument definition (relevant code). I believe the idea is to let the module have sane behavior depending on changing conditions and not have to commit to any particular way of doing things. Quoting from the docs: Whether to override the default patch merge approach with a specific type. By the default, the strategic merge will typically be used.

Ok let's have it same as they have.

PR #114, opened 2018-10-16
Fix wait codepath in KubeVirtVM and add tests for it

mmazur:
1. Enable it. Was ommited before.
2. Make sure return values are serializable, otherwise there are errors when jsonizing on module exit.

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
@@ -178,6 +178,11 @@ class KubeVirtVM(KubernetesRawModule):
         argspec.update(copy.deepcopy(RAW_ARG_SPEC))
         return argspec
 
+    def fix_serialization(self, obj):
+        if obj and hasattr(obj, 'to_dict'):
+            return obj.to_dict()
+        return obj
+
     def execute_module(self):
         changed = False
         results = []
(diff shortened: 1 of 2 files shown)


lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
```diff
@@ -237,11 +243,14 @@ def _read_stream(self, resource, watcher, stream):
                     if (not self._use_cdi(annotations, labels) or
```
machacekondra: please change `IMPORT_STATUS_KEY` to ```IMPORT_STATUS_KEY = 'cdi.kubevirt.io/storage.pod.phase'```

mmazur: Will do.

mmazur: Btw, what is the rationale for this? When I dump the object returned by the stream on PVC creation, the relevant parts look like this:

```
    "metadata": {
      "annotations": {
        "pv.kubernetes.io/bind-completed": "yes", 
        "pv.kubernetes.io/bound-by-controller": "yes"
      }, 
```

There's no _cdi.kubevirt.io_ of any kind to be found. Is it because we're using different cluster versions or something?

machacekondra: That's what's used by the cdi:

https://github.com/kubevirt/containerized-data-importer/blob/9ba4059dfc6657adcf27e2a112273d087d0aa12a/pkg/controller/controller.go#L23

I guess the previous name is just a typo.

mmazur: ok

Issue #79, opened 2018-08-01
Rename kubevirt_raw module


tripledes:
Following the lead of kubernetes ansible modules, it'd be good idea to rename `kubevirt_raw` module to just `kubevirt`, as it seems the `raw`  portion of it scared potential users as they took it as unfinished or not something for the general public.

lizsurette:
+1 to this. I had an additional comment on the naming of the modules that I wanted to add here. I'm happy to create a new issue if this shouldn't live here though :)

For clarity and consistency with other files/actions should we rename “kubevirt_raw_pvc.yml” to “kubevirt_pvc_create.yml”?

mmazur:
Ever since the move to the dynamic api client for kubernetes, the `kubevirt_raw` module was on its way out, as in effect it became just a copy of the upstream `k8s` module. We're currently in the process of replacing `kubevirt_raw` with dedicated kubevirt and k8s modules. Once that's done, `kubevirt_raw` should be dropped to not cause confusion among users.

PR #115, opened 2018-10-24
Add kubevirt_vm module

machacekondra:
This PR is work in progress, and kind of propsal of simplyfing design to the `kubevirt_vm` module.

Currently only `memory`, `disks` and `volumes` are simplified. to be first level instead of nested. As per current implementeation I would keep the same interface of API as kubevirt is using so we will keep the diff/dry mode functionality of k8s module.

Also I propse to have just single module `kubevirt_vm` instead of two `kubevirt_raw` and `kubevirt_vm_status`. Feel free to comment.

```yaml
    - name: Create VM
      kubevirt_vm:
        state: present
        name: testvm
        namespace: default
        memory: 64M
        disks:
         - name: registrydisk
           volumeName: registryvolume
           disk:
             bus: virtio
         - name: cloudinitdisk
           volumeName: cloudinitvolume
           disk:
             bus: virtio
        volumes:
         - name: registryvolume
           registryDisk:
             image: kubevirt/cirros-registry-disk-demo
         - name: cloudinitvolume
           cloudInitNoCloud:
             userDataBase64: SGkuXG4=
```

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
@@ -1,0 +1,388 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2018, KubeVirt Team <@kubevirt>
+# Apache License, Version 2.0
+# (see LICENSE or http://www.apache.org/licenses/LICENSE-2.0)
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: kubevirt_vm
+
+short_description: Manage KubeVirt virtual machine
+
+description:
+    - Use Openshift Python SDK to manage the state of KubeVirt virtual machines.
+
+version_added: "2.8"
+
+author: KubeVirt Team (@kubevirt)
+
+options:
+    state:
+        description:
+            - Set the virtual machine to either I(present), I(absent), I(running) or I(stopped).
+            - "I(present) - Create or update virtual machine."
+            - "I(absent) - Removes virtual machine."
+            - "I(running) - Create or update virtual machine and run it."
+            - "I(stopped) - Stops the virtual machine."
+        default: "present"
+        choices:
+            - present
+            - absent
+            - running
+            - stopped
+        type: str
+    name:
+        description:
+            - Name of the virtual machine.
+        required: true
+        type: str
+    namespace:
+        description:
+            - Namespace where the virtual machine exists.
+        required: true
+        type: str
+    disks:
+        description:
+            - List of dictionaries which specify disks of the virtual machine.
+            - A disk can be made accessible via four different types: I(disk), I(lun), I(cdrom), I(floppy).
+            - https://kubevirt.io/api-reference/master/definitions.html#_v1_disk
+        type: list
+    volumes:
+        description:
+            - List of volumes specification for the virtual machine.
+            - https://kubevirt.io/api-reference/master/definitions.html#_v1_volume
+        type: str
+    labels:
+        description:
+            - Labels are key/value pairs that are attached to virtual machines. Labels are intended to be used to
+              specify identifying attributes of virtual machines that are meaningful and relevant to users, but do not directly
+              imply semantics to the core system. Labels can be used to organize and to select subsets of virtual machines.
+              Labels can be attached to virtual machines at creation time and subsequently added and modified at any time.
+            - More on labels that are used for internal implementation:
+              https://kubevirt.io/user-guide/#/misc/annotations_and_labels
+        type: dict
+    machine_type:
+        description:
+            - QEMU machine type is the actual chipset of the virtual machine.
+        type: str
+    ephemeral:
+        description:
+            - If (true) ephemeral vitual machine will be created. When destroyed it won't be accessible again.
+            - Works only with C(state) I(present) and I(absent).
+        type: bool
+        default: false
+
+extends_documentation_fragment:
+  - k8s_auth_options
+  - k8s_resource_options
+
+requirements:
+  - python >= 2.7
+  - openshift >= 0.6.2
+'''
+
+EXAMPLES = '''
+- name: Start virtual machine 'myvm'
+  kubevirt_vm:
+      state: running
+      name: myvm
+      namespace: vms
+
+- name: Create virtual machine 'myvm' and start it
+  kubevirt_vm:
+      state: running
+      name: myvm
+      namespace: vms
+      memory: 64M
+      disks:
+        - name: registrydisk
+          volumeName: registryvolume
+          disk:
+            bus: virtio
+
+- name: Combine inline definition with Ansible parameters:
+  kubevirt_vm:
+      # Kubernetes specification:
+      definition:
+        metadata:
+          labels:
+            app: galaxy
+            service: web
+            origin: vmware
+
+      # Ansible parameters:
+      state: running
+      name: myvm
+      namespace: vms
+      memory: 64M
+      disks:
+        - name: registrydisk
+          volumeName: registryvolume
+          disk:
+            bus: virtio
+        - name: cloudinitdisk
+          volumeName: cloudinitvolume
+          disk:
+            bus: virtio
+      volumes:
+        - name: registryvolume
+          registryDisk:
+            image: kubevirt/cirros-registry-disk-demo:latest
+        - name: cloudinitvolume
+          cloudInitNoCloud:
+
+- name: Start ephemeral virtual machine 'myvm' and wait to be running
+  kubevirt_vm:
+      ephemeral: true
+      state: running
+      wait: true
+      wait_time: 20
+      name: myvm
+      namespace: vms
+      memory: 64M
+      labels:
+        kubevirt.io/vm: myvm
+      disks:
+        - name: registrydisk
+          volumeName: registryvolume
+          disk:
+            bus: virtio
+        - name: cloudinitdisk
+          volumeName: cloudinitvolume
+          disk:
+            bus: virtio
+      volumes:
+        - name: registryvolume
+          registryDisk:
+            image: kubevirt/cirros-registry-disk-demo:latest
+        - name: cloudinitvolume
+          cloudInitNoCloud:
+            userDataBase64: IyEvYmluL3NoCgplY2hvICdwcmludGVkIGZyb20gY2xvdWQtaW5pdCB1c2VyZGF0YScK
+
+- name: Remove virtual machine 'myvm'
+  kubevirt_vm:
+      state: absent
+      name: myvm
+      namespace: vms
+'''
+
+RETURN = '''
+vm:
+    description:
+      - The virtual machine that user manages.
+      - It contains all attributes https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachine
+    returned: success
+    type: complex
+'''
+
+import copy
+import traceback
+
+from collections import defaultdict
+
+from ansible.module_utils.k8s.common import AUTH_ARG_SPEC, COMMON_ARG_SPEC
+from ansible.module_utils.k8s.raw import KubernetesRawModule
+
+from openshift import watch
+from openshift.dynamic.client import ResourceInstance
+from openshift.helper.exceptions import KubernetesException
+
+
+VM_ARG_SPEC = {
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']},
+    'ephemeral': {'type': 'bool', 'default': False},
+    'state': {
+        'type': 'str',
+        'choices': [
+            'present', 'absent', 'running', 'stopped'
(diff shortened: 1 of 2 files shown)


machacekondra:
@pkliczewski @mmazur @masayag @vatsalparekh @adityaramteke

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -0,0 +1,312 @@
...
+        type: str
+    name:
+        description:
+            - Name of the virtual machine.
+        required: true
+        type: str
+    namespace:
+        description:
+            - Namespace where the virtual machine exists.
+        required: true
```
pkliczewski: I would do it optional. We could use `default` if not provided.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -0,0 +1,312 @@
...
+version_added: "2.8"
+
+author: KubeVirt Team (@kubevirt)
+
+options:
+    state:
+        description:
+            - Set the virtual machine to either C(present), C(absent), C(running) or C(stopped).
+        default: "present"
+        choices:
```
pkliczewski: Would be great to have a doc which explains what each state means.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -0,0 +1,312 @@
...
+        type: str
+    namespace:
+        description:
+            - Namespace where the virtual machine exists.
+        required: true
+        type: str
+    disks:
+        description:
+            - List of dictionaries which specify disks of the virtual machine.
+            - A disk can be made accessible via four different types: I(disk), I(lun), I(cdrom), I(floppy).
```
pkliczewski: As far as I can see we do not provide such information when defining disks. Please see [here](https://github.com/kubevirt/kubevirt/blob/master/cluster/examples/vmi-fedora.yaml#L11).

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -0,0 +1,312 @@
...
+            - https://kubevirt.io/api-reference/master/definitions.html#_v1_disk
+        type: list
+    volumes:
```
pkliczewski: Would be great to specify additional information such as registryDisk, pvc etc.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -0,0 +1,312 @@
...
+        description:
+            - Volume specification for the virtual machine.
+            - https://kubevirt.io/api-reference/master/definitions.html#_v1_volume
+        type: str
+    persistent:
```
pkliczewski: Why do we need this? How do we map that to vm/vmi definition?

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -0,0 +1,312 @@
...
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']},
+    'persistent': {'type': 'bool', 'default': True},
+    'state': {
+        'type': 'str',
+        'choices': [
+            'present', 'absent', 'running', 'stopped'
+        ],
+        'default': 'present'
+    },
+    'wait': {'type': 'bool', 'default': True},
```
pkliczewski: Please use wait, wait_time in one of examples.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -0,0 +1,312 @@
...
+}
+
+
+class KubeVirtVM(KubernetesRawModule):
+    def __init__(self, *args, **kwargs):
+        super(KubeVirtVM, self).__init__(*args, **kwargs)
+
+    @property
+    def argspec(self):
+        """ argspec property builder """
```
pkliczewski: I would like to keep the old code so we can still support `inline`. In this case it would be great to share the code.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm_status.py
```diff
@@ -1,184 +0,0 @@
-#!/usr/bin/python
-# -*- coding: utf-8 -*-
-
-# Copyright (c) 2018, KubeVirt Team <@kubevirt>
-# Apache License, Version 2.0
-# (see LICENSE or http://www.apache.org/licenses/LICENSE-2.0)
-
-ANSIBLE_METADATA = {'metadata_version': '1.1',
```
pkliczewski: I would keep this code so we could still support inline in case user want to use something which is not supported.

pkliczewski:
I really like the approach. Let's discuss the details :)

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: Well, I think it's not good idea, the reason I did not do it even in oVirt modules is that if you for example have a VM for example `testvm` in namespace `testns`, and you want to update that vm memory lets say, if you forget to specify the namespace you may change vm in namespace `default` which is something you don't want, and with current aprroach it will fail. But we can discuss.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: Sure, will add.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: Not sure I follow. You mean that disk type don't have to be specified in case we want to remove disk?

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: In case user do following:

```yaml
kubevirt_vm:
  name: myvm
  state: running
  memory: 64M
   disks:
     - ...
```

With `persistent=true/false` he can decide if he want to create VM object and also VMI, or he want to create only VMI object without VM object.

machacekondra: The case you didn't understand from documentation means, it's wrong documentation, I will try to improve.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: Ok

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: Yes it will be still supported.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
pkliczewski: Sounds reasonable

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
pkliczewski: I meant that we do not specify things like luns, cdroms etc.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
pkliczewski: I think user can decide by creating specific object VM or VMI. Not sure about the value of this flag.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
pkliczewski: cool, would be great to have an example of each.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: Yes, but where do he specify it?

machacekondra: Actually I think we should hide this kind of implementation detail from him, so it's more user friendly to work with the module for him.

pkliczewski: You are right it is better to hide this implementation detail. I would consult the name with UX and let's keep it.

machacekondra:
Fixes: https://github.com/kubevirt/ansible-kubevirt-modules/issues/79

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
mmazur: 1. The arg description is missing info that this defaults to True.
2. I'm not sure 'persistent' is the best term for this, though I can't really think of a better one. (Ephemeral? But that's usually used for storage, not vms.)
3. The description should contain short wording describing what implementation detail is being hidden here, so that people who actually understand how kubevirt operates don't have to guess what this does. Something like: _Persistent VMs create a VirtualMachine object in the cluster and a matching VirtualMachineInstance whenever state is /running/. Non–persistent VMs create only the latter object, which gets irrecoverably deleted whenever state is set to anything other than /running/._

mmazur:
Hmm, I have an idea. Instead of permanent: true/false, let's just have a vm_type:
1. Permanent singleton. (Better name needed. :) aka VM
2. Ephemeral aka VMI
3. Replicated aka VMIReplicaSet.

That way one sets up/controls all the vms with a single module with basically the same syntax.

(And maybe the kubevirt inventory plugin should adhere to the same abstraction.)

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
pkliczewski: Seems like `ephemeral` is good name for this. I would update the documentation stating it is `False` be default.

pkliczewski:
@mmazur I think that `vm_type` is too much detail exposed. I would hide this detail and used the same code and pass `kind` in the implementation to distinguish it.

mmazur:
If we don't want to explicitly expose the internal resource names, then I'd suggest the following:
- _ephemeral_: bool; default **false**
  - _true_ – create VMI
  - _false_ – create VM
- _replicas_: int; default **-1**
  - _<0_ – create VM or VMI
  - _>=0_ – create VMIReplicaSet

The module should reject ephemeral=true && replicas>=0, since that's not valid. Otherwise we can control all currently existing resource types with these two arguments.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: yeah `ephemeral` is used on many places in kubevirt doc, so would be the best to use it as well

machacekondra:
> Though I'm still of the opinion that vm_type: persistent, ephemeral, replicated, preset is a better approach.

`empheral` seems ok argument to me to have in `kubevirt_vm` module.

About `replica` argument. I've read `VirtualMachineInstanceReplicaSet` documentation. And it seems to me as nice paramater to `kubevirt_vm` as well.

`Preset` doesn't suite that much. If we would add it as some parameter, then I would call the module `kubevirt_vm`, but just `kubevirt` as it manages everything. Maybe we can create generic enough code, and create separate module for `presets` and maybe even for `replica_sets`.

pkliczewski:
I agree with @machacekondra let's work on `vm` related code and try to make it as generic as possible. Later we could add support for `preset`s or `replica_set`s.

mmazur:
Agreed then. Let's ignore presets completely and for this PR start with ephemeral yes/no. We can add replica support later as a separate argument.

This approach should btw inform how the inventory plugin is designed. Even if we're only interested in VMIs, then adding an indication whether they're part of a VM (or a replicaset) is trivial by checking for the existence of vm.metadata.ownerReferences and verifying kind == VirtualMachine.

machacekondra:
What's missing:

 - Wating for stopped state.
 - Empheral vm don't support wait and works only with state present and absnet

I think it's OK to be reviewed/merged, if you don't insist on those two missing things.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -0,0 +1,390 @@
...
+              imply semantics to the core system. Labels can be used to organize and to select subsets of virtual machines.
+              Labels can be attached to virtual machines at creation time and subsequently added and modified at any time. 
+            - More on labels that are used for internal implementation:
+              https://kubevirt.io/user-guide/#/misc/annotations_and_labels
+        type: dict
+    machine_type:
+        description:
+            - QEMU machine type is the actual chipset of the virtual machine.
+        type: str
+    empheral:
```
mmazur: _ephemeral_ – this typo is in all instances of this word.

machacekondra:
Fixed, but not added tests. I try to add them later.

PR #117, opened 2018-10-24
Expand test coverage for scaling module

mmazur:
This puts us at 76% total test coverage according to pytest and thus completes the __improve unit tests__ task. I hope…


pkliczewski:
@mmazur What about pvc and wait logic there?

mmazur:
PVC's wait logic was covered by PR #114. This one takes care of the wait logic for scaling.

Also, I have no idea what to do about the travis checks failing. If I run the same commands on my own system, they pass without issues.

mwperina:
Maybe travis is using different pyflake version than you. Anyway according to log, this is the error which is being raised:

```
lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py:244:29: W504 line break after binary operator
```

pkliczewski:
@mmazur Yeah, @mwperina is right. We need travis to be happy.

mmazur:
Got it to 80%, ready for merge.

PR #118, opened 2018-10-25
Unify (some) resource names in sample playbooks

mmazur:
Those playbooks are useful for testing, but only if they operate
on the same resources.


pkliczewski:
@mmazur Let's make sure Travis is happy.

mmazur:
@pkliczewski yes, and you can do that by merging #119. That will also "fix" all the other PRs.

Issue #116, opened 2018-10-24
Scaling module assumes VirtualMachineInstanceReplicaSet values that aren't there


mmazur:
According to https://kubevirt.io/docs/workloads/controllers/virtual-machine-replica-set.html (last paragraph), a return object from a VMIRS invocation should contain "status/ready replicas" values. That's also how kubernetes' built–in RS does it.

The upstream k8s scaling code assumes __object.status.readyReplicas__ to exist as does our code ([here](https://github.com/kubevirt/ansible-kubevirt-modules/blob/master/lib/ansible/modules/clustering/kubevirt/kubevirt_scale_vmirs.py#L220)). Trouble is – VMIRS does not return that, so using scaling with __wait__ always fails with a timeout, since the code simply requires readyReplicas to be there.

I don't know what to do about this.

pkliczewski:
> I don't know what to do about this.

Let's check what the code returns and fix our code so it can work with VMIRS.

mmazur:
The issue is that our code is correct, it's just that VMIRS doesn't return the values required for our code to work. So I can't really fix it in our code.

What I'll do is prepare a simple test case for this and have all of you run it on your envs. If it works for you, then it's something with my env. If it doesn't, then we have an upstream bug and I'll fill a case with upstream.

mmazur:
Martin suggested to just go upstream with this. I'll check it against latest dev env and write up my problem in an issue against kubevirt itself.

pkliczewski:
@mmazur If you confirm that it doesn't work with latest master please open issue on KubeVirt and paste it here as reference.

mmazur:
I've checked today with latest kubevirt and provider k8s-1.11.0 and emailed kubevirt-dev asking if anyone would like to clarify whether it's supposed to be this way.

mmazur:
(Potential) discussion, for reference: https://groups.google.com/forum/#!topic/kubevirt-dev/WLoWkkvFgOE

mmazur:
Roman helped me debug, it was an issue with my env.

PR #120, opened 2018-10-25
[WIP] Fix scaling wait code for current VMIRS behavior

mmazur:
This makes the scaling work.

This is regarding issue #116. Assuming upstream VMIRS behavior is correct, this code is mergable as–is. I've emailed kubevirt-dev and am waiting for clarification on this. If they say current behavior is fine, then this patch fixes the last bug I've stumbled upon.


pkliczewski:
@mmazur Let's wait on Travis to be happy and combine this with the tests.

mmazur:
Ok, I've added the tests to this PR. We should now wait until, say, Monday, for anybody to respond to my email regarding this. If it gets ignored or the current behavior is confirmed as correct, then PR #117 can be dropped and this one merged instead.

mmazur:
Ok, I've bumped the coverage slightly, but now this PR requires #123 to be merged first.

mmazur:
Bug in my env, this patch is wrong.

PR #123, opened 2018-10-26
Also test failure codepaths

mmazur:
Intro PR for bumping the test coverage by a few percent.


mmazur:
This would be the next one to apply.

PR #125, opened 2018-10-30
Clean up VMIPreset sample playbooks

mmazur:
The idea is that kubevirt_raw_vmi.yml and kubevirt_raw_vmi_from_preset.yml create 1:1 same VMIs, they just do it slightly differently.


pkliczewski:
@vatsalparekh @machacekondra

PR #127, opened 2018-10-31
Fix facts playbooks and rename them to k8s_facts

mmazur:
Those playbooks didn't get tested with k8s_facts properly (they don't work). And they should've been renamed to clearly indicate we're using upstream code for facts gathering.


mmazur:
@machacekondra keep in mind that with the new rules in place, if you approve a change to non–core code, you should merge it as well. (I'll add a few sentences to the readme on which part is core and which isn't on Friday.)

PR #128, opened 2018-11-06
Add contributing guidelines

mmazur:
@machacekondra @pkliczewski – do these sound reasonable to you? I'm hoping they'd speed up work by making some stuff explicit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,9 @@
+# Contributing Guidelines
+
+## Merging Pull Requests
+
+1. Don't merge your own PRs.
+2. Changes that touch core code (`lib/`) or its unit tests (`tests/units/`) require at least two positive reviews.
+3. New code (e.g. new modules) also requires at least two positive reviews.
+4. Other changes, especially small ones, can be merged with just a single positive review.
+5. If you've done the review and are allowed to merge according to the rules, but for whatever reason don't feel comfortable doing so, it's perfectly fine to ask for another opinion. (Though make sure to explicitly ask someone else. Don't just comment "looks good to me" and close the browser tab.)


README.md
```diff
@@ -122,3 +122,15 @@ $ ansible-playbook tests/playbooks/kubevirt_vm_facts.yml
 ```
 
 The above command, will gather the information for the VM stated in the playbook and print out a JSON document based on [KubeVirt VM spec](https://kubevirt.io/api-reference/master/definitions.html#_v1_virtualmachine).
+
+
+## Development
+
+### Rules for merging pull requests
```
vatsalparekh:
```suggestion
### Guidelines for merging pull requests
```

machacekondra:
Looks OK to me. Just thinking if this information should be in README. I think 'normal' user is not insterested in this, only developers, maybe creating README_developer.md or CONTRIBUTING.md, or anything like that?

mmazur:
Yup, contributing or something sounds good.

README.md
```diff
@@ -122,3 +122,15 @@ $ ansible-playbook tests/playbooks/kubevirt_vm_facts.yml
...
+
+1. Don't merge your own PRs.
+2. Changes that touch core code (`lib/`) or its unit tests (`tests/units/`) require at least two approvals.
+3. New code (e.g. new modules) also requires at least two approvals.
+4. If a PR has gotten the required 2+ approvals while being a work in progress (marked by a `[WIP]` prefix in the title) and it's clear what the required changes still are, then once those final changes get pushed by the author, it's ok for a single reviewer to approve and merge such a PR. In other words – you don't need to wait for a second reviewer before merging a PR, if a recent enough version of said PR has already been approved by someone else. Just make sure the final changes are what was agreed upon, then merge.
```
pkliczewski: I think we should not approve something which is not finished. We can review it but acks should be done on something that author considers to be ready.

pkliczewski:
Like Ondra mentioned we should have this in `CONTRIBUTING.md` since this approach other projects are using.

README.md
mmazur: Hmm. That's a bit of an issue – I did not intend „approvals” to mean the github feature „Kowalski approves this change”, but a more general term referring to somebody saying something like „looks good to me”. Basically what happened with the new kubevirt_vm module:
1. We've had discussions (both here and with videochat) discussing the change.
2. We basically agreed it just needs some minor work to be mergeable.
3. Ondra pushed final version, I looked it over, got him to fix the last few stoppers, then merged it.

Would it be ok if I just changed „approvals” to „positive reviews” everywhere? Technically the only actual approval needed (?) would be by the person doing the merge.

pkliczewski: We can agree on the approach but the review needs to be done for "ready" PR. We should not approve something which is not finished. In the text you mentioned approvals but with Ondra's change we only commented not approved. Above text is not correct even for the PR you refer as an example.

masayag: @pkliczewski will it be possible to enable labels ? Instead of relying on comments, labels will serve as a cleared indicator of the PR status (i.e. wip / review needed / 2 reviews needed )

pkliczewski: @masayag Good idea! Labels to the rescue :muscle:

mmazur: Hmm, though if we started to use labels, then we'd always have the option to agree beforehand while working on a WIP that, for example, it's almost finished and we can drop the requirements to one more review == merge (and relabel). That way we get the same speedup, while making sure we're all on board with such a decision… yeah, guess that could work as well. Ok, I'll drop the paragraph and add something about labels.

mmazur:
Ok, I've added _reviews/1_ and _reviews/2_ labels and removed the WIP paragraph from the merge rules. It should now be easier to track mergeability of PRs and make those decisions.

PR #129, opened 2018-11-07
Improve merge dict so it works with nested dict

machacekondra:
This patch fixes the `merge_dics` method. Previously it didn't work with nested dictionaries. For example merging dictinaries like:

```
dict1 = {
  'labels': {'label1': 'value'}
}
dict2 = {
  'labels': {'label2': 'value'}
}
```

Ended as `  'labels': {'label1': 'value'}` instead of `  'labels': {'label1': 'value', 'label2': 'value'}`.

This patch fixes it.


machacekondra:
@mmazur @pkliczewski

pkliczewski:
@mmazur please review

mmazur:
Looks good to me.

PR #130, opened 2018-11-07
Add sample playbook with a simple service definition

mmazur:
One example should be enough – it's just to show that there are no differences between VMIs and standard container pods.


pkliczewski:
@mmazur The question is what user would need to provide:
* name
* service type
* ports
* selector
Everything else could be set as defaults.

mmazur:
Ok, I'll try to figure out whether there's a useful way to do this.

PR #131, opened 2018-11-08
kubevirt_vm: Move watch operation code into check_mode statement

imjoey:
Within the method `_manage_state()`, the `watch` related operations should probably be ignore when the `check_mode` is `true`.

Note that `check_mode` has been checked via https://github.com/kubevirt/ansible-kubevirt-modules/blob/master/lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py#L368  beforehand. So maybe we could also remove the `check_mode` condition check codes in `_manage_state()`. Any thoughts on this?

Signed-off-by: imjoey <majunjiev@gmail.com>


machacekondra:
@imjoey Great to see contributing you to the kubevirt Ansible ! :)

imjoey:
hi @machacekondra , so nice to meet you, here. 😄  I recently notice this amazing project kubevirt and I believe it will later play such an important role in cloud computing. 😄

imjoey:
@machacekondra  That's reasonable. Done. Thanks.

PR #134, opened 2018-11-09
Unify configs in sample playbooks; fix kubevirt account

mmazur:
Took the cloudinit from kubevirt examples. Short and usable for testing.

(The previous one didn't really work and I actually use our playbooks for dev work.)


tests/test_rsa
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
pkliczewski: What is the value of changing key file name?

tests/playbooks/kubevirt_raw_vmirs.yml
```diff
@@ -30,29 +30,23 @@
...
-                         gecos: KubeVirt Project
-                         sudo: ALL=(ALL) NOPASSWD:ALL
-                         passwd: $6$JXbc3063IJir.e5h$ypMlYScNMlUtvQ8Il1ldZi/mat7wXTiRioGx6TQmJjTVMandKqr.jJfe99.QckyfH/JJ.OdvLb5/OrCa8ftLr.
-                         shell: /bin/bash
-                         home: /home/kubevirt
-                         lock_passwd: false
-                         ssh_authorized_keys:
-                             - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC+KMp1F4SrKitvgILf8Q2GQbcOslIg4sfqUOf+m832zwoS2Z2ndmpr/3F+NtLq56cTaEzK9IPBIkNLJfsPqRsToLBYZD+l4S5FfuZ2xwCxGBH+TVXApO+SiD6c84rmjOx47665iQvMHKL+n/5gVvSdYDuegNKnj4rRr/eHnG2yC4TVZl3oHI7TPOUJT+kKjSWP1UesWTZmmck39IaFSmorg31X7g9hJHwq9JEDQilcbnIyqDZKiH6Ju4GjOU8mqhazBFB4qu/QDDb25pDpPd2pQGBilGvm7gwJCVnyDk9YZQU7gYE734KLDf5tCKMmEQSjwFx2Tj9mfZveCIJkaj3T kubevirt
-                 name: cloudinitvolume
+                     password: fedora
```
pkliczewski: We lost a bit of configuration. What is the intention to delete it?

tests/playbooks/kubevirt_raw_vm.yml
```diff
@@ -29,30 +29,23 @@
...
-                         shell: /bin/bash
-                         home: /home/kubevirt
-                         lock_passwd: false
-                         ssh_authorized_keys:
-                             - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC+KMp1F4SrKitvgILf8Q2GQbcOslIg4sfqUOf+m832zwoS2Z2ndmpr/3F+NtLq56cTaEzK9IPBIkNLJfsPqRsToLBYZD+l4S5FfuZ2xwCxGBH+TVXApO+SiD6c84rmjOx47665iQvMHKL+n/5gVvSdYDuegNKnj4rRr/eHnG2yC4TVZl3oHI7TPOUJT+kKjSWP1UesWTZmmck39IaFSmorg31X7g9hJHwq9JEDQilcbnIyqDZKiH6Ju4GjOU8mqhazBFB4qu/QDDb25pDpPd2pQGBilGvm7gwJCVnyDk9YZQU7gYE734KLDf5tCKMmEQSjwFx2Tj9mfZveCIJkaj3T kubevirt
-                 name: cloudinitvolume
+                     password: fedora
+                     chpasswd: { expire: False }
+                     ssh_authorized_keys:
+                       - ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDizcn50tdTi4TiaUP3tuyLKQSgXmAMXkF/s8pNb+2agcudaxtwsz+2I9MH8Ty9eXDTMg3b5kqzNQRtuNyG8QX3Lm+Q0M2rTIPVxe6ZkVbdgXzSPQ0Lx8GUBu/AF0qv40sC2Rmbu1vbc2ogSHwkj/yDnGPoMNiVeG3Z6ZGFRwuLznU+C0owILOGgf6m83gNhALomLmNxsov+y+vUh5ofyEW0ve21htpsM3G/ui4/8hXVK8P92By84jzNkTCv6yTIPbpZ3XUr9AGTt5YgfGHMBAoyJ1rEsjjroDeGn0pVhIxeqsTP1/ej//hKwDvpMlbcrsWN6lAg0qfv4ikNf7OE7/N test@kubevirt
```
machacekondra: Why not use here the ` tests/test_rsa.pub`?

tests/test_rsa
mmazur: 1. I find it a good rule to name any publicized ssh key pairs as 'test' or 'example' or 'sample' to make sure it's clear this isn't meant to be used for stuff.
2. I'm no longer creating a 'kubevirt' user, so don't want the key named 'kubevirt'.

tests/playbooks/kubevirt_raw_vmirs.yml
mmazur: It didn't work and figuring out why would not be a good use of my time. I checked what kubevirt examples do, verified that that worked and switched to using that.

tests/test_rsa
pkliczewski: and what is the reason for not creating the user.

tests/playbooks/kubevirt_raw_vmirs.yml
pkliczewski: Before doing any changes we need to understand why something is not working.

tests/playbooks/kubevirt_raw_vm.yml
mmazur: You mean lookup('file', …)? Good point, I'll change it in a sec.

tests/playbooks/kubevirt_raw_vmirs.yml
mmazur: No, not really. Figuring out how fedora 27 parses cloud init configs is extremely out of scope for what we're supposed to be doing. If you want to add debugging this as a task to trello – sure. But if not, then this is just a minor roadblock for the task I'm actually supposed to be doing and I'd rather not spend any more time on it. It already took me long enough to get it working (e.g. turns out latest ssh-keygen generates private keys that older ssh clients don't parse… because reasons).

pkliczewski: In my opinion we need to understand the issue first so let's base our assumptions on the documentation. It is not ok to say I do not understand the code so I am going to change it.

mmazur: The difference between the code that was and the new code is: test user is called 'fedora' instead of 'kubevirt'. It does not seem reasonable to me to spend time on figuring out how to switch it back in lieu of continuing with my assigned task.

mmazur: Hmm, I think I wasn't clear on this two messages up: cloud-init configs are passed basically literally to the vm and then parsed by scripts in the 'cloud-init' package on linux. So me figuring out how to get a 'kubevirt' user account running would have zero to do with kubevirt or ansible or anything we're supposed to be doing. It'd be as arbitrary as adding, say, 'dnf install vim htop' as an initialization step. End result is I'd get admin_skill++ and that's it – our stuff would in no way be "better" due to such a change.

```
[root@test-fedora-vmi mnt]# cat /mnt/user-data
#cloud-config
password: fedora
chpasswd: { expire: False }
ssh_authorized_keys:
  - "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDizcn50tdTi4TiaUP3tuyLKQSgXmAMXkF/s8pNb+2agcudaxtwsz+2I9MH8Ty9eXDTMg3b5kqzNQRtuNyG8QX3Lm+Q0M2rTIPVxe6ZkVbdgXzSPQ0Lx8GUBu/AF0qv40sC2Rmbu1vbc2ogSHwkj/yDnGPoMNiVeG3Z6ZGFRwuLznU+C0owILOGgf6m83gNhALomLmNxsov+y+vUh5ofyEW0ve21htpsM3G/ui4/8hXVK8P92By84jzNkTCv6yTIPbpZ3XUr9AGTt5YgfGHMBAoyJ1rEsjjroDeGn0pVhIxeqsTP1/ej//hKwDvpMlbcrsWN6lAg0qfv4ikNf7OE7/N test@kubevirt"```

pkliczewski: The trick is that we do not have good documentation yet and those playbooks are used as examples. I still do not see good reason why not to keep it. Here is the [documentation](https://cloudinit.readthedocs.io/en/latest/topics/examples.html) which should help to understand it.

pkliczewski:
@mmazur Please remove `WIP` if you consider it to be ready.

mmazur:
Ok, the fixes applied against the `kubevirt` user seem to work. This is mergeable now. While not useful long–term (because `kubevirt_raw`), the cloud-init config in those playbooks should be the basis for similar playbooks utilizing `kubevirt_vm`, so that vms created from the new playbooks can actually be used logged onto and used for testing/experimenting.

PR #135, opened 2018-11-14
kubvirt_vm: Add support for networks & interfaces

machacekondra:
This PR propose structure to `interfaces` and `networks` parameters. I think the best solution would be to just pass it to kubevirt API again, but just flatten the structure. I was thinking even about doc, and again I propose just adding a link to official doc + some light doc of the parameter, and add example.

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
@@ -51,12 +51,12 @@ options:
         description:
             - List of dictionaries which specify disks of the virtual machine.
             - A disk can be made accessible via four different types: I(disk), I(lun), I(cdrom), I(floppy).
-            - https://kubevirt.io/api-reference/master/definitions.html#_v1_disk
+            - All possible configuration options are available in U(https://kubevirt.io/api-reference/master/definitions.html#_v1_disk)
         type: list
     volumes:
         description:
             - List of volumes specification for the virtual machine.
-            - https://kubevirt.io/api-reference/master/definitions.html#_v1_volume
+            - All possible configuration options are available in U(https://kubevirt.io/api-reference/master/definitions.html#_v1_volume)
         type: str
     labels:
         description:
@@ -64,9 +64,15 @@ options:
               specify identifying attributes of virtual machines that are meaningful and relevant to users, but do not directly
               imply semantics to the core system. Labels can be used to organize and to select subsets of virtual machines.
               Labels can be attached to virtual machines at creation time and subsequently added and modified at any time.
-            - More on labels that are used for internal implementation:
-              https://kubevirt.io/user-guide/#/misc/annotations_and_labels
+            - More on labels that are used for internal implementation U(https://kubevirt.io/user-guide/#/misc/annotations_and_labels)
         type: dict
+    interfaces:
+        description:
+            - An interface defines a virtual network interface of a virtual machine (also called a frontend).
+            - All possible configuration options interfaces are available in U(https://kubevirt.io/api-reference/master/definitions.html#_v1_interface)
+            - Each interface must have specified a I(network) that declares which logical or physical device it is connected to (also called as backend).
+              All possible configuration options of network are available in U(https://kubevirt.io/api-reference/master/definitions.html#_v1_network).
+        type: list
     machine_type:
         description:
             - QEMU machine type is the actual chipset of the virtual machine.
@@ -105,6 +111,22 @@ EXAMPLES = '''
           volumeName: registryvolume
           disk:
             bus: virtio
+
+- name: Create virtual machine 'myvm' with multus network interface
+  kubevirt_vm:
+      name: myvm
+      namespace: vms
+      memory: 512M
+      interfaces:
+        - name: default
+          bridge: {}
+          network:
+            pod: {}
+        - name: mynet
+          bridge: {}
+          network:
+            multus:
+              networkName: mynetconf
 
 - name: Combine inline definition with Ansible parameters:
   kubevirt_vm:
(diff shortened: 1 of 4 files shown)


machacekondra:
@mmazur @pkliczewski @vatsalparekh @adityaramteke This is just a proposal, feel free to comment, if you disagree with something.

machacekondra:
Or we could create just single parameter `network` and as subparameter add interface, maybe something like:

```
kubevirt_vm:
  ...
  networks:
     - name: x
        interface:
          bridgde: {}
        multus:
          networkName: mynetconf
```

machacekondra:
Or other way around:

```
kubevirt_vm:
  ...
  interfaces:
    - name: x
       bridge: {}
       network: 
         multus:
           networkName: mynetconf
```

Actually this would be closest to oVirt IMHO, which is:

```
ovirt_vm:
  ...
  nics:
    - name: X
      inteface: virtio
      profile_name: mgmtbridge
```

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -51,22 +51,32 @@
...
               imply semantics to the core system. Labels can be used to organize and to select subsets of virtual machines.
               Labels can be attached to virtual machines at creation time and subsequently added and modified at any time.
-            - More on labels that are used for internal implementation:
-              https://kubevirt.io/user-guide/#/misc/annotations_and_labels
+            - More on labels that are used for internal implementation U(https://kubevirt.io/user-guide/#/misc/annotations_and_labels)
         type: dict
+    interfaces:
+        description:
+            - An interface defines a virtual network interface of a virtual machine (also called a frontend).
+            - Each interface must have a corresponding C(network) with the same name.
```
pkliczewski: Is `same name` required by KubeVirt or is it our simplification?

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -106,6 +116,23 @@
           disk:
             bus: virtio
 
+- name: Create virtual machine 'myvm' with multus network interface
+  kubevirt_vm:
+      name: myvm
+      namespace: vms
+      memory: 512M
+      interfaces:
```
pkliczewski: Is there a way to specific static network configuration such as IP or MAC addresses?

tests/playbooks/kubevirt_vm_multus.yml
```diff
@@ -0,0 +1,65 @@
+- name: Create Fedora vm with two network interfaces
+  hosts: localhost
+  gather_facts: false
+  connection: local
+
+  # Before running this playbook you need to make sure you have created
+  # 'mynetconf' NetworkAttachmentDefinition. This is used by the second
```
pkliczewski: Should we provide a link to documentation how to do it?

pkliczewski:
@dankenigsberg Please review or assign someone who will do it.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: In current approach yes, but if we would go with following design, it wouldnt:

```
kubevirt_vm:
  ...
  interfaces:
    - name: x
       bridge: {}
       network: 
         multus:
           networkName: mynetconf
```

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: Not any I am aware of, it would be possible in case of using multus and OVS, if you pass cloud_init specyfing the IP, but deffinitelly not by passing it in interface,

tests/playbooks/kubevirt_vm_multus.yml
machacekondra: The only we have is following:

 https://kubevirt.io/user-guide/#/workloads/virtual-machines/interfaces-and-networks?id=multus

I will paste it.

pkliczewski: OK

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
pkliczewski: OK

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
pkliczewski: Sounds good.

machacekondra: Ok, in that case I would do the same also for volumes and disks.

machacekondra:
Here is running example: https://asciinema.org/a/212176

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -106,6 +112,22 @@
...
+        - name: default
+          bridge: {}
```
mmazur: Everything looks fine except this part – if it's fine to have an empty `bridge` parameter, then why don't we just not require specifying it in the config? Let's have it just default to {} and not have the user think about it.

machacekondra: For now I would merge as-is and think of general solution for handling reasonable nested defaults.

pkliczewski: @mmazur It is ok with you?

mmazur: Sure, this is a minor issue.

Though if we have an agreed upon list of improvements, where do we put it? Maybe in the module itself as a comment?

pkliczewski: Or an issue which could be handled by next PR.

machacekondra: https://github.com/kubevirt/ansible-kubevirt-modules/issues/137

PR #136, opened 2018-11-15
Add a dedicated module for managing Services

mmazur:
Let's make sure we're on the same page wrt to config parameters early. @machacekondra @pkliczewski 

The idea behind `spec` is:
1. One can completely ignore the module's parameters and provide the whole Service spec inline.
2. Any parameter not covered directly by the module for whatever reason can always be provided this way.

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/k8s_service.py
+++ b/lib/ansible/modules/clustering/kubevirt/k8s_service.py
@@ -1,0 +1,205 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2018, KubeVirt Team <@kubevirt>
+# Apache License, Version 2.0
+# (see LICENSE or http://www.apache.org/licenses/LICENSE-2.0)
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: k8s_service
+
+short_description: Manage Services on Kubernetes
+
+description:
+    - Use Openshift Python SDK to manage Services on Kubernetes
+
+version_added: "2.8"
+
+author: KubeVirt Team (@kubevirt)
+
+options:
+    name:
+        description:
+            - Use to specify a Service object name.
+        required: true
+        type: str
+    namespace:
+        description:
+            - Use to specify a Service object namespace.
+        required: true
+        type: str
+    type:
+        description:
+            - Specifies the type of Service to create.
+            - U(https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types)
+        choices:
+            - NodePort
+            - ClusterIP
+            - LoadBalancer
+            - ExternalName
+        default: ClusterIP
+        required: true
+    ports:
+        description:
+            - A list of ports to expose.
+            - U(https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services)
+        required: true
+        type: list
+    selector:
+        description:
+            - Label selectors identify objects this Service should apply to.
+            - U(https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/)
+        required: true
+        type: dict
+
+extends_documentation_fragment:
+  - k8s_auth_options
+  - k8s_resource_options
+
+requirements:
+  - python >= 2.7
+  - openshift >= 0.6.2
+'''
+
+EXAMPLES = '''
+- name: Expose https port with ClusterIP
+  k8s_service:
+    state: present
+    name: test-https
+    namespace: default
+    type: ClusterIP
+    ports:
+    - port: 443
+      protocol: TCP
+    selector:
+      key: special
+'''
+
+RETURN = '''
+result:
+  description:
+  - The created, patched, or otherwise present Service object. Will be empty in the case of a deletion.
+  returned: success
+  type: complex
+  contains:
+     api_version:
+       description: The versioned schema of this representation of an object.
+       returned: success
+       type: str
+     kind:
+       description: Always 'Service'.
+       returned: success
+       type: str
+     metadata:
+       description: Standard object metadata. Includes name, namespace, annotations, labels, etc.
+       returned: success
+       type: complex
+     spec:
+       description: Specific attributes of the object. Will vary based on the I(api_version) and I(kind).
+       returned: success
+       type: complex
+     status:
+       description: Current status details for the object.
+       returned: success
+       type: complex
+     items:
+       description: Returned only when multiple yaml documents are passed to src or resource_definition
+       returned: when resource_definition or src contains list of objects
+       type: list
+'''
+
+import copy
+import traceback
+
+from collections import defaultdict
+
+from ansible.module_utils.k8s.common import AUTH_ARG_SPEC, COMMON_ARG_SPEC
+from ansible.module_utils.k8s.raw import KubernetesRawModule
+
+
+SERVICE_ARG_SPEC = {
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']},
+    'selector': {'type': 'dict'},
+    'type': {
+        'type': 'str',
+        'choices': [
+            'NodePort', 'ClusterIP', 'LoadBalancer', 'ExternalName'
+        ],
+        'default': 'ClusterIP',
+    },
+    'ports': {'type': 'list'},
+}
+
+API_VERSION = 'v1'
+
+
+def virtdict():
+    return defaultdict(virtdict)
+
+
+class KubernetesService(KubernetesRawModule):
+    def __init__(self, *args, **kwargs):
+        super(KubernetesService, self).__init__(*args, **kwargs)
+
+    @staticmethod
+    def merge_dicts(x, y):
+        for k in set(x.keys()).union(y.keys()):
+            if k in x and k in y:
+                if isinstance(x[k], dict) and isinstance(y[k], dict):
+                    yield (k, dict(KubernetesService.merge_dicts(x[k], y[k])))
+                else:
+                    yield (k, y[k])
+            elif k in x:
+                yield (k, x[k])
+            else:
+                yield (k, y[k])
+
+    @property
+    def argspec(self):
+        """ argspec property builder """
+        argument_spec = copy.deepcopy(COMMON_ARG_SPEC)
+        argument_spec.update(copy.deepcopy(AUTH_ARG_SPEC))
+        argument_spec.update(SERVICE_ARG_SPEC)
+        return argument_spec
+
+    def execute_module(self):
+        """ Module execution """
+        self.client = self.get_api_client()
+
+        selector = self.params.get('selector')
+        service_type = self.params.get('type')
+        ports = self.params.get('ports')
+
+        definition = virtdict()
+
+        def_spec = definition['spec']
+        def_spec['type'] = service_type
+        def_spec['ports'] = ports
+        def_spec['selector'] = selector
+
+        # Override with 'resource_definition:' if provided
+        definition = dict(self.merge_dicts(definition, self.resource_definitions[0]))
+
+        resource = self.find_resource('Service', self.api_version, fail=True)
+        definition = self.set_defaults(resource, definition)
+        result = self.perform_action(resource, definition)
+
+        self.exit_json(**result)
+
+
+def main():
+    module = KubernetesService()
+    try:
+        module.api_version = API_VERSION
+        module.execute_module()
+    except Exception as e:
+        module.fail_json(msg=str(e), exception=traceback.format_exc())
+
+
+if __name__ == '__main__':
+    main()
--- a/tests/playbooks/k8s_service.yml
+++ b/tests/playbooks/k8s_service.yml
@@ -4,19 +4,19 @@
   connection: local
   tasks:
     - name: Expose ssh port on test-fedora-vmi
-      k8s:
+      k8s_service:
         state: present
+        name: test-fedora-vmi-ssh-access
+        namespace: default
+        type: NodePort
+        ports:
+          - port: 22
+            protocol: TCP
+        selector:
+          kubevirt/vmi: test-fedora-vmi
         definition:
-          apiVersion: v1
-          kind: Service
           metadata:
-            name: test-fedora-vmi-ssh-access
-            namespace: default
+            labels:
+              test: label
           spec:
-            type: NodePort
-            ports:
-              - port: 22
-                nodePort: 30022
-                protocol: TCP
-            selector:
-              name: test-fedora-vmi
+            externalTrafficPolicy: Cluster


machacekondra:
Few things we should discuss.

1) The name `kubevirt_service`, it actually manages k8s service, so I think we should rename.
2) Maybe shouldn't be this code part of `kubevirt_vm`?

mmazur:
@machacekondra well, yeah, there would be absolutely nothing kubevirt–specific about the module. So k8s_service sounds more like it. On the other hand it'd be kind of weird, since we're not the upstream module. Not sure what to think about it.

Regarding integrating this with kubevirt_vm: that was my initial plan, though considering how complicated it'd be to implement inside kubevirt_vm, my feeling is that it'd make sense only if it was limited in what it could do – I was thinking only ClusterIP, since to me that's the most natural thing for a vm – you define a vm at the same time and configure all the ports you want to make available on some publicly–accessible IP.

But the task is to support everything Services can do, so Piotr said that I should just do the separate module and maybe pitch the kubevirt_vm integration to Moran, since he'd know whether consumers of this would be interested in stuff like it. So I mailed Moran – we'll see how that goes.

So tl;dr – the initial implementation should be a separate module and the extension for kubevirt_vm might be done later.

mmazur:
Oh, and one more thing. Having seen your implementation of multus, basic Service integration would be a perfect extension for it. Since we have to define the primary default k8s–handled interface anyway (all pods have to have it), I'd just add syntax to define port mappings for that particular interface. The logic would be:
1. If only one 'default' interface is defined, then just create the Service with the port mappings and don't try to use multus at all.
2. If multiple interfaces defined, then create Service for the first one and configure everything else as multi–nic.

It'd be a perfect simplified syntax from the pov of kubevirt_vm's consumers.

lib/ansible/modules/clustering/kubevirt/kubevirt_service.py
```diff
@@ -0,0 +1,78 @@
...
+        type: str
+    namespace:
+        description:
+            - Use to specify a Service object namespace.
+        required: true
+        type: str
+    selector:
+        description:
+            - TODO
+        type: dict
```
pkliczewski: Are we going to support k8s semantics for selectors here?

lib/ansible/modules/clustering/kubevirt/kubevirt_service.py
```diff
@@ -0,0 +1,78 @@
...
+        choices:
+            - NodePort
+            - ClusterIP
+            - LoadBalancer
+    ports:
+        description:
+            - TODO
+        required: true
+        type: list
+    labels:
```
pkliczewski: I would go with default here `[]`

lib/ansible/modules/clustering/kubevirt/kubevirt_service.py
mmazur: Yup, no other way to do it.

lib/ansible/modules/clustering/kubevirt/kubevirt_service.py
mmazur: Of course.

machacekondra: I think you thought `{}`. But also we must be careful with this as now if there is a service existing and you run this module and don't define any label, it will remove all labels from that service.

machacekondra:
@mmazur I don't see any issue with naming it k8s_service and sending it upstream, if they would like to accept it. 

Separate module is perfectly OK, as we will have the role to encapsulate the logic into single unit anyway. I was just saying that we should consider it as well, if it's not anything too complex and makes sense to do.

Yep, we can have limited support for it, and more complex support in separate module. Makes sense to me.

lib/ansible/modules/clustering/kubevirt/kubevirt_service.py
pkliczewski: Yeah, I saw list above but indeed I meant dictionary.

mmazur: @machacekondra well yes, but what other options do we have? If my definition has for example two labels, then an update adds those labels. If I remove them from my yaml and then update, then all labels get cleared. I don't think we're supposed to support a case where k8s objects are partially maintained by our ansible module and partially by something else. Or do I not know something about how k8s modules are supposed to work?

mmazur:
Ok, I've renamed the it to k8s_service – we can see check if upstream would want something like this.

Seems to work. Now need to fill out the missing parts of the docs and actually test all the options.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,208 @@
...
+# Apache License, Version 2.0
+# (see LICENSE or http://www.apache.org/licenses/LICENSE-2.0)
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: kubevirt_service
```
machacekondra: rename also here

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,208 @@
...
+        type: str
+    namespace:
+        description:
+            - Use to specify a Service object namespace.
+        required: true
+        type: str
+    selector:
+        description:
+            - TODO
+        default: []
```
machacekondra: {}

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,208 @@
...
+        description:
+            - TODO
+        required: true
+        type: list
+    labels:
+        description:
+            - Labels are key/value pairs that are attached to objects.
+        default: {}
+        type: dict
+    spec:
```
machacekondra: I think it's called ` resource_definition ` and aliases definition, inline.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,208 @@
...
+
+import copy
+import traceback
+
+from collections import defaultdict
+
+from ansible.module_utils.k8s.common import AUTH_ARG_SPEC, COMMON_ARG_SPEC
+from ansible.module_utils.k8s.raw import KubernetesRawModule
+
+from openshift import watch
```
machacekondra: I think you don't need this.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,208 @@
...
+        definition = dict(self.merge_dicts(self.resource_definitions[0], definition))
+
+        resource = self.find_resource('Service', self.api_version, fail=True)
+        definition = self.set_defaults(resource, definition)
+        result = self.perform_action(resource, definition)
+        changed = result['changed']
+
+        self.exit_json(**{
+            'changed': changed,
+            'service': result,
```
machacekondra: Not sure if it's part of the result, but I think we should definitelly return user-friendly IP address, based on service type

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,208 @@
...
+def virtdict():
+    return defaultdict(virtdict)
+
+
+class KubernetesService(KubernetesRawModule):
+    def __init__(self, *args, **kwargs):
+        super(KubernetesService, self).__init__(*args, **kwargs)
+
+    @staticmethod
+    def merge_dicts(x, y):
```
pkliczewski: We already have this method [here](https://github.com/kubevirt/ansible-kubevirt-modules/blob/dce8831abbe04174d098ee00c0a0dbd61a701cdf/lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py#L228). Please reuse.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,208 @@
...
+        argument_spec = copy.deepcopy(COMMON_ARG_SPEC)
+        argument_spec.update(copy.deepcopy(AUTH_ARG_SPEC))
+        argument_spec.update(SERVICE_ARG_SPEC)
+        return argument_spec
+
+    def execute_module(self):
+        """ Module execution """
+        self.client = self.get_api_client()
+
+        selector = self.params.get('selector')
```
pkliczewski: What about `name` and `namespace`?

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: They're not the same thing. `resource_definition` in upstream k8s encompasses the whole object passed to the cluster. In my code `spec` only extends/overrides the `spec:` part of a Service definition. Naming it anything other than `spec` would imply that it does something other than what it actually does.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: That's a good general question. For something like PVC creation or replicaset modification it makes sense to have a --wait options, since volume space allocation/new instance creation might take time. But there's no way this makes sense for a Service module, right?

lib/ansible/modules/clustering/kubevirt/k8s_service.py
machacekondra: Ok, sounds reasonable, but in that case we need to always support everything which isn't part of spec.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: If upstream rejects a merge request, then I'll separate this into common code. But with the goal of  submitting this upstream, I can't extract this method (yet).

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Most of the heavy lifting is done by KubernetesRawModule. Including dealing with the always obligatory name/ns/apiver/kind parameters.

pkliczewski: It that is the case we can remove it from other modules as well.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
pkliczewski: There is no need to copy this code. It seems like good reason to create a utility class.

lib/ansible/modules/clustering/kubevirt/kubevirt_service.py
machacekondra: What I just wanted to say is that having `None` as default in this case is more safe IMHO.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
machacekondra: With this comment I meant you don't use it now. But to your question, I am unsure how services works. the question is what it should wait for ... The IP address to appear ?

pkliczewski: The real question is whether we need the address at all.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: I don't know if that's doable in case of our older modules. Even if it is, it'd count more as a rewrite of the module's internals. We'd be rewriting working code.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: I can't create a `module_utils/kubevirt/utils.py`, have `k8s_service.py` depend on its contents and then send a PR adding `k8s_service.py` to upstream ansible. They'd reject it with _there's code missing here_. _If_ they reject the PR and we end up keeping `k8s_service.py`_ in our own repo, _then_ I can have it use kubevirt's utility code. Not before.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: That's an interesting problem. The solutions are:
1. Add `metadata:`, analogous to `spec`, to circumvent the issue.
2. Go through the v1 api and ensure I support all the metadata fields in existence.

Though if I go do the latter, I might as well add all the `spec` fields in existence as module options and drop `spec:` entirely. It did cross my mind, though I suspect that might end up a mine field.

Yup: I've added `labels`, but forgot for example `annotations`. I think it might make sense to just drop `labels:` and go with the `metadata:` approach. Then we're covered.

Though to be safe I'll try to dig up a full schema for Services, brb.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Yeah, I was hoping you'd know… ;) Anyway, as is tradition, I guess we can start with the initial version of this module not supporting --wait, and then figure out if we want/need it later.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
machacekondra: I think full schema is better as with your current solution, you would need to maintain new parameters still watch if there is any new etc. With full schema you just don't care.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
machacekondra: Users creates service just because of address, no? Or what's the use case of creating service without the need of address?

pkliczewski: ATM, I do not see a need for IP so I think there is no need to wait.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
pkliczewski: I am ok with doing it in separate PR. In general please do not copy any code.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
machacekondra: Ah right, in case of nodeport. Ok, then. Thanks.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Yeah, I need to add `metadata:` (and drop `labels:`), since there's [just too much of it](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#objectmeta-v1-meta) plus there's likely to be more with newer k8s releases. I should probably also add `metadata:` to `kubevirt_vm`.

`spec:` is also [rather large](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#servicespec-v1-core), which makes me think it's pointless to try to map all the arguments and instead the goal should be to support the required/most often used arguments and leave everything else in `spec:`.

machacekondra: No need to add it to `kubevirt_vm` `metadata` is supported by `resource_definition` already.

mmazur: I've just pushed a version that drops this entirely and relies on `definition:` for any additional metadata (like labels).

mmazur:
Ok, the current push is basically finished. Needs unit tests, but at present there's no more functionality work left.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Good thing I looked at this. We need to change the way we run `exit_json()`. Currently we produce `result.result.*`. Now I've changed it to just `result.*`.

Back to your question. Currently one gets that info with `result.spec.clusterIP`. I don't really see the point of adding a second name for that value. Let's stick with what k8s does if it looks clear enough.

lib/ansible/modules/clustering/kubevirt/kubevirt_service.py
mmazur: Dropped `labels:` entirely, so no longer an issue.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Running the code as is results with the clusterIP getting assigned without having to wait for anything. So if one wants to use that info in the playbook, it's available as `result.spec.clusterIP`.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
machacekondra: Ok,  I am fine if it's nested.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,199 @@
...
+        description:
+            - Specifies the type of Service to create.
+        choices:
+            - NodePort
+            - ClusterIP
+            - LoadBalancer
+        required: true
+    ports:
+        description:
+            - A list of ports to expose.
```
machacekondra: Well user don't know if it's list of strings, int, dicts.. Can you please paste here url to doc what could be provided here?

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,199 @@
...
+from ansible.module_utils.k8s.raw import KubernetesRawModule
+
+
+SERVICE_ARG_SPEC = {
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']},
+    'selector': {'type': 'dict', 'default': {}},
+    'type': {
+        'type': 'str',
+        'choices': [
+            'NodePort', 'ClusterIP', 'LoadBalancer'
```
machacekondra: The sad thing is that since now on, we need to maintain this list. But I think this list won't be changed to often, so it's ok

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,199 @@
...
+            - LoadBalancer
+        required: true
+    ports:
+        description:
+            - A list of ports to expose.
+        required: true
+        type: list
+    selector:
+        description:
+            - Label selectors identify objects this Service should apply to.
```
pkliczewski: I think that here we could benefit from having a more information about what we can provide here and what are the types.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,199 @@
...
+        default: {}
```
pkliczewski: I wonder whether it makes sense to create a service without selector `{}`. I think we should require it every time. Documentation suggest not to use selectors in:
> * You want to have an external database cluster in production, but in test you use your own databases.
> * You want to point your service to a service in another Namespace or on another cluster.
> * You are migrating your workload to Kubernetes and some of your backends run outside of Kubernetes.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
pkliczewski: In K8s documentation we have [one more type](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) which is ignored by KubeVirt. We need to maintain this list with any supported type.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Yeah, I was wondering how to handle this. I think you're right, we should go with requiring a selector and if someone really really knows what they're doing, they can set an empty selector via definition.metadata.selector (that should work).

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Yeah, for upstreaming this is a must. Added.

mmazur:
Ok, I've addressed the points of discussions we've had. This is the most simple version of this that we can have that gets us the functionality we need. Please merge. I'll put the unit tests in a separate PR and continue any further work on this with upstream.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,203 @@
...
+SERVICE_ARG_SPEC = {
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']},
+    'selector': {'type': 'dict', 'default': {}},
+    'type': {
+        'type': 'str',
+        'choices': [
+            'NodePort', 'ClusterIP', 'LoadBalancer', 'ExternalName'
+        ],
+    },
+    'ports': {'type': 'list', 'default': []},
```
machacekondra: Remeber that those default will always reset already existing services.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,203 @@
...
+
+from collections import defaultdict
+
+from ansible.module_utils.k8s.common import AUTH_ARG_SPEC, COMMON_ARG_SPEC
+from ansible.module_utils.k8s.raw import KubernetesRawModule
+
+
+SERVICE_ARG_SPEC = {
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']},
+    'selector': {'type': 'dict', 'default': {}},
```
machacekondra: Remeber that those default will always reset already existing services.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Just to make sure we're understanding each other. If I have a config like this:

```
type: ClusterIP
ports:
  - port: 22
selector:
  key: special
```

And for whatever reason want to keep the clusterip, but stop exposing any ports (doesn't make sense in this case, but we might be doing other modules where it will):

```
type: ClusterIP
selector:
  key: special
```

then yes, I'd expect the Service in the cluster to match 1:1 the playbook, which in this case means: delete the ports if they're no longer in the playbook.

Is that not correct behavior?

machacekondra: Yes this will do exactly what you described, but what if you want to update selector, and you don't want to specify all the current details of the service, but just want to define selector and name of the service, because it's easier. You will always have to define ports, otherwise you will change the setup.

mmazur: I went through k8s docs and k8s ansible module code and you are most likely correct that not enforcing this the way I did is the right way to do it. I'll push the changes tomorrow morning.

mmazur: Ok, pushed the patch without those defaults.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,203 @@
...
+        type: str
+    namespace:
+        description:
+            - Use to specify a Service object namespace.
+        required: true
+        type: str
+    type:
+        description:
+            - Specifies the type of Service to create.
+            - https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
```
machacekondra: you can use U(https://url) tag, to make it real link

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,203 @@
...
+
+version_added: "2.8"
+
+author: KubeVirt Team (@kubevirt)
+
+options:
+    name:
+        description:
+            - Use to specify a Service object name.
+        required: true
```
machacekondra: This is documented as required, but not really required if I see correctly.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,203 @@
...
+    type:
+        description:
+            - Specifies the type of Service to create.
+            - https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
+        choices:
+            - NodePort
+            - ClusterIP
+            - LoadBalancer
+            - ExternalName
+        required: true
```
machacekondra: This is documented as required, but not really required if I see correctly.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,203 @@
...
+    ports:
+        description:
+            - A list of ports to expose.
+            - https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services
+        required: true
```
machacekondra: This is documented as required, but not really required if I see correctly.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -0,0 +1,203 @@
...
+        type: list
+    selector:
+        description:
+            - Label selectors identify objects this Service should apply to.
+            - https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
+        required: true
```
machacekondra: This is documented as required, but not really required if I see correctly.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
machacekondra: Also I would be here more precise, this is what k8s doc says:
```
type determines how the Service is exposed. Defaults to ClusterIP. Valid options are ExternalName, ClusterIP, NodePort, and LoadBalancer. "ExternalName" maps to the specified externalName. "ClusterIP" allocates a cluster-internal IP address for load-balancing to endpoints. Endpoints are determined by the selector or if that is not specified, by manual construction of an Endpoints object. If clusterIP is "None", no virtual IP is allocated and the endpoints are published as a set of endpoints rather than a stable IP. "NodePort" builds on ClusterIP and allocates a port on every node which routes to the clusterIP. "LoadBalancer" builds on NodePort and creates an external load-balancer (if supported in the current cloud) which routes to the clusterIP. More info: https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services---service-types
```
Maybe at least add link to that doc?

lib/ansible/modules/clustering/kubevirt/k8s_service.py
machacekondra: I think adding (also?) a doc to api specification would be good.

machacekondra:
If the module won't be accepted upstream, the doc isn't important anyway...

mmazur:
I've gotten the first round of comments on the upstream PR. There's a good chance it'll get accepted. Please merge as–is and I'll continue work on the upstream PR (including your doc suggestions).

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Yes, but considering that there are only very specific cases where you'd want to not have a selector, it's probably a better choice from a UX perspective to have you set it as empty if you really know you don't want it.

machacekondra: I am ok whether it's required or not. What I mean is that doc says something different than the code, if I see correctly.

PR #138, opened 2018-11-20
Added CRUD tests for kind

vatsalparekh:
CRUD test for Kinds in KubeVirt, supposed to be also suitable to run as gate for new PR


tests/playbooks/kind_crud.yml
```diff
@@ -0,0 +1,172 @@
...
+        state: absent
+        namespace: default
+        name: test-fedora
+        inline:
+         apiVersion: kubevirt.io/v1alpha2
+         kind: VirtualMachine
+         metadata:
+           name: test-fedora
+           namespace: default
+      register: vm_facts
```
machacekondra: Here you should test if it reported changed.

vatsalparekh: ack, added

pkliczewski:
@vatsalparekh What about update? I do not see that recreate is changing anything.

vatsalparekh:
@pkliczewski added update part, except VMI which cannot be updated.

vatsalparekh:
@machacekondra @pkliczewski @mmazur @adityaramteke PTAL

tests/playbooks/kind_crud.yml
```diff
@@ -0,0 +1,735 @@
...
+  
+    - name: Fail if VM create task was not 'changed'
+      assert:
+        that:
+        - "vm_facts is changed"
+        fail_msg: "Create VM task did not return changed"
+    
+    - name: wait for changes to appear
+      wait_for: timeout=15
```
pkliczewski: Timeout is with seconds. Can we wait on object to change like we had in import vm apb:

> - Name: Check PVC bound status
>  command: "oc get pvc {{ vm_name }}-disk-01 -o jsonpath='{ {{pvc_status_path}} }'"
>  register: pvc_bound_status
>  until: pvc_bound_status.stdout == 'Bound'
>  retries: 60
>  delay: 3

tests/playbooks/kind_crud.yml
```diff
@@ -0,0 +1,735 @@
...
+      k8s_service:
+        state: present
+        name: test-fedora-vmi-ssh-access
+        namespace: default
+        type: NodePort
+        ports:
+          - port: 22
+            protocol: TCP
+        selector:
+          kubevirt/vmi: test-fedora-vmi
```
pkliczewski: I think it should be `kubevirt/vm` here. Can you check whether it works with `vmi` as well?

vatsalparekh: @pkliczewski IIRC, we usually expose VMI using `virtctl`
it resembles with other example as well
https://github.com/kubevirt/ansible-kubevirt-modules/blob/21b6a857a395f841c359d80ea78ee61366109178/tests/playbooks/k8s_service.yml#L16

pkliczewski: Hmm, this is interesting. I used example command from kubevirt [doc](https://kubevirt.io/user-guide/#/workloads/virtual-machines/expose-service):

> virtctl expose virtualmachineinstance vmi-ephemeral --name vmiservice --port 27017 --target-port 22

and when described created service I see `Selector:          kubevirt.io/vm=example-box`

tests/playbooks/kind_crud.yml
```diff
@@ -0,0 +1,756 @@
...
+      register: vm_facts
+  
+    - name: Fail if VM create task was not 'changed'
+      assert:
+        that:
+        - "vm_facts is changed"
+        fail_msg: "Create VM task did not return changed"
+    
+    - name: wait for changes to appear
+      command: "oc get vm test-fedora -o jsonpath='{.spec.running}'"
```
machacekondra: using k8s facts, would be better here imho

tests/playbooks/kind_crud.yml
```diff
@@ -0,0 +1,756 @@
...
+
+    - name: Fail if VM recreate task was 'changed'
+      assert:
+        that:
+        - "vm_facts is not changed"
+        - "vm_facts is success"
+        fail_msg: "Create VM task did return changed"
+  
+    - name: shutdown test-fedora VM
+      kubevirt_vm_status:
```
machacekondra: Please use `kubevirt_vm` module

tests/playbooks/kind_crud.yml
```diff
@@ -0,0 +1,756 @@
...
+    
+    - name: wait for changes to appear
+      command: "oc get vm test-fedora -o jsonpath='{.spec.running}'"
+      register: vm_facts
+      until: vm_facts.stdout == 'false'
+      retries: 60
+      delay: 3
+
+    - name: Power on test-fedora VM
+      kubevirt_vm_status:
```
machacekondra: Please use kubevirt_vm module

mmazur:
Is there any point in merging this if it uses `kubevirt_raw`? Shouldn't we have all of it use the Newer Better modules?

machacekondra:
> Is there any point in merging this if it uses kubevirt_raw? Shouldn't we have all of it use the Newer Better modules?

I agree, @vatsalparekh can you please migrate to new `kubevirt_vm` module. If you need any help with that, just ping me.

vatsalparekh:
@mmazur @machacekondra @pkliczewski 
The reason for using `kubevirt_vm_status` and `kubevirt_raw` along side with the new `kubevirt_vm` was to have coverage of all the modules we have, not that `kubevirt_vm` has any limitations. This playbook is supposed to serve as test.

machacekondra:
Ok, let's use it for now. I will send a patch removing old stuff.

PR #139, opened 2018-11-20
kubevirt_vm: structure returned k8s objects differently

mmazur:


lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -376,10 +376,7 @@ def execute_module(self):
                 ret = self.manage_state(state)
                 changed = changed or ret
```
pkliczewski: We still populate changed param. Why do we want to remove from `exit_json` and leave it in the code?

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -376,10 +376,7 @@ def execute_module(self):
...
 
-        self.exit_json(**{
-            'changed': changed,
-            'vm': result,
-        })
+        self.exit_json(**result)
```
machacekondra: The reason wer return it like this is because we return proper changed also if running state was changed..

mmazur: Wait, so if you change `spec.running` from _true_ to _false_ or vice versa, then the `result` from `perform_action()` will not have `changed: true`? Why?

pkliczewski: so maybe instead of removing it we should fix the bug you mentioned above.

mmazur: Ah, now I've read the code. If you're running `patch_resource()` inside of `manage_state()`, then I think we should just override the `changed` value from `result` instead of embedding `result` one level down. It's not like that value is somehow incorrect.

mmazur: I've just pushed an (untested) version of what I've had in mind as the better solution to this.

mmazur: Yup, tested, works as I've intended. I believe this is a better approach of reporting results from this module. And in line with what upstream k8s and our other modules do.

machacekondra: So please also update RETURN documentation appropriatelly

mmazur: Ok, copied the k8s_service RETURN docs, since that's exactly what we'll be returning as well.

machacekondra: Yeah and now comes questions if we want to, I've created 'vm' return type, to hide implementation detail of VMI/VM, now user gets always different things. `items` are not supported here, I've always use first `resource_definition`. I am fine with this return type, but this *must* be backward comaptible so we must decide once and forever.

mmazur: Yup, you're right on your reasoning. I think I can get this to a state where we'll both be happy. I'll post a patch soonish.

machacekondra: Well, this is simplification module, we must remember that. Most of the ansible modules return whatever it manages. For example see here:
  https://docs.ansible.com/ansible/latest/modules/stat_module.html#return-values

I think we really should just return 'vm' with all of it API attributes. So I believe we agree on this.

For the `k8s_objects`, not sure I follow. I think as part of vm dict we should return all vm attributes. But just a single VM. I think we shouldn't support multiple VM management in this module, but I am open changing this, if there is good reason, and not too much maintanence effort.

So IMHO, we should just return vm dict and inside it the vm attributes.

machacekondra: even more likely, maybe we should define the return value `kubevirt_vm` not just `vm`.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -174,11 +174,16 @@
...
+  description:
+    - The virtual machine managed by the user.
+  returned: success
+  type: complex
+  contains:
+    k8s_objects:
+      description:
+      - A list of kubernetes objects this module created, modified or read during operation.
+      returned: success
+      type: list
```
pkliczewski: There is one object in the list. How can we get more than one object?

mmazur: Yes. I didn't have time to get it completely finished yesterday, but the idea is this:
1. Once I finish the patch, it'll return a VM and potentially a VMI when operating on one.
2. If we ever add, say, built–in port mapping handling, then it also might return a Service object.
3. We've discussed supporting kubevirt longterm without having to change the module too much. This approach helps accomplish such a task, since if we ever need to support any new resource types, they just get returned as part of `k8s_objects`.

pkliczewski: > 1. Once I finish the patch, it'll return a VM and potentially a VMI when operating on one.

The question is why do we want to expose both so far we did good job in hiding `VMI` implementation details. What is the value for a user?

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
mmazur: The point of the module is to abstract away the lowlevel stuff and make it seem like this is just a vm. Currently this means we're operating on both VMs and VMIs, yet returning only the former's results with some modifications (`changed`). In the future we might end up handling even more resource types and/or different versions of VMs/VMIs, which will return yet more different things. I don't think we should just assume that VM is the thing we'll be returning and what the user should expect to be getting.

I think that if we really want to be safe long term, we should ensure that the abstraction is properly handled on the RETURN values. So my idea is that we put the abstracted results directly under (following your suggestion) `kubevirt_vm` complex return structure and put any raw k8s objects we've touched during operation inside the `kubevirt_vm.k8s_objects` list. So the end form of RETURN might be something like this:

- `kubevirt_vm.primaryIP` – empty if stopped, taken from VMI if running.
- `kubevirt_vm.publicIP` – if we ever support ClusterIP Service, then that'd be the clusterIP; empty if not configured
- `kubevirt_vm.apiVersion` – if there are multiple VM versions, then that'd probably be it
- (…) – who knows what we'll get requests for as being useful to end users
- `kubevirt_vm.k8s_objects` – a list of low–level objects that we've used for those users that actually know their kubevirt and need to access a very specific VMI parameter, for example.

Granted, I'm not fluent enough in ansible to know how return values are used in practice. Maybe `k8s_objects` should be a dict and not a list, so that one can do `kubevirt_vm.k8s_objects.VMI.result.xyz`, but that implies we always operate on a single resource type – makes sense for now, but I'm not 100% certain it'll hold true for ever. The point is figuring out a way to return *everything* that might be potentially useful to end users.

Anyway, that's my reasoning for this change in the context of getting long term stable and flexible RETURN interface.

mmazur: tl;dr – I don't like the discrepancy between abstracting away VMs/VMIs on the input, but then neither actually abstracting it fully nor returning a full complement of low–level objects on the output. Instead I'd like us to do both.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
mmazur: Long term I'm afraid that if we commit to pretending only VMs exist despite that not being the case, we might end up regretting it. Full reasoning is in the other thread.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: Ok now I understand what you've meant by `k8s_objects`. It makes sense. Just note, those most probably will be `kubevirt_objects`, but that's just cosmetic note.

Well, we can return it, no problem with that. I am just worried it would be too complicated.

mmazur: Putting my admin hat on: best software is designed in such a way, that doing basic stuff is trivial, yet arbitrarily complicated stuff is still achievable. Worst times as an admin are where the designer thought „there's *surely* no way anyone would need to do *that*”. Hence my approach to achieving both: curated high level return values should we need them and an easy access to all the raw low level results for the 1%.

That said, I've talked myself into switching from a list to a dict. If we ever need to return, say, multiple VMIs in `k8s_objects`, then we can always have lists like `k8s_objects.VMIlist` or `k8s_objects.VMIs`. I'll push a modified version in a bit.

mmazur:
Ok, since Ondra would like stuff merged, this is the version I'd like to consider for inclusion. Basically it's just a bunch of renames, but one that allows me to also expose VMI cleanly in a later patch.

mmazur:
And tests first…

mmazur:
And ready.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -195,12 +195,17 @@
...
-    type: complex
+kubevirt_vm:
+  description:
+    - The virtual machine managed by the user.
+  returned: success
+  type: complex
+  contains:
+    k8s_objects:
+      description:
+      - Lists kubernetes objects this module created, modified or read during operation, by type name.
```
machacekondra: It doesn't currently return, the VMI. But I wouldn't hold this PR anymore, let's merge and create issue to fix it.

PR #140, opened 2018-11-20
Added expose and ssh for e2e playbook

adityaramteke:


tests/playbooks/e2e.yaml
```diff
@@ -117,3 +120,21 @@
...
+      k8s_service:
+        state: present
+        name: test-fedora-vmi
+        namespace: default
+        type: NodePort
+        ports:
+          - port: 20022
+            protocol: TCP
+        selector:
+          kubevirt.io/vm: test-working
```
pkliczewski: I do not see `test-working` anywhere. I see that everywhere we are using vm name as parameter ` '{{ vm_name }}'`. I think we should do it here as well.

adityaramteke: Yes, I did those changes need to push.

tests/playbooks/e2e.yaml
```diff
@@ -93,7 +94,29 @@
         name: '{{ vm_name }}'
         namespace: default
 
-    - name: Incrase the resources of VM
+    - name: Wait for VMI status to running
```
machacekondra: You don't have to. The `kubevirt_vm` with `status=running` will wait for you,

tests/playbooks/e2e.yaml
```diff
@@ -93,7 +94,29 @@
...
+      retries: 18
+
+    - name: Wait for VM to be ready
+      k8s_facts:
+        name: "{{ vm_name }}"
+        namespace: default
+        kind: VirtualMachine
+        api_version: kubevirt.io/v1alpha2
+      register: vm
+      until: "vm.resources[0].status['created'] == true and vm.resources[0].status['ready'] == true"
```
machacekondra: Hmm, we don't have this in wait for running state in `kubevirt_vm` module, I think we should add it.

tests/playbooks/e2e.yaml
adityaramteke: @machacekondra okay, should I remove it?

machacekondra: Yes please. If it's removed it will then properly test if the module is waiting.

tests/playbooks/e2e.yaml
adityaramteke: @machacekondra are you going to add this to kubevirt_vm or should I keep it?

machacekondra: Keep it for now, I will remove it once we add it to `kubevirt_vm`.

PR #141, opened 2018-11-20
kubevirt_raw: correct exit_json for multiple results

mmazur:
This: results[5].result.spec.running
is better than this: result.results[5].result.spec.running

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
@@ -211,9 +211,7 @@ class KubeVirtVM(KubernetesRawModule):
 
         self.exit_json(**{
             'changed': changed,
-            'result': {
-                'results': results
-            }
+            'results': results
         })
 
     def _create_stream(self, resource, namespace, wait_time):


lib/ansible/modules/clustering/kubevirt/kubevirt_raw.py
```diff
@@ -211,9 +211,7 @@ def execute_module(self):
 
         self.exit_json(**{
             'changed': changed,
-            'result': {
-                'results': results
-            }
+            'results': results
```
machacekondra: Please update RETURN doc

machacekondra: Just document it's `resutls` and it's list of the complex types, which are already documented.

machacekondra: I think k8s has it same

machacekondra:
Shouldn't we actually drop this module?

mmazur:
We should. I'm closing this, there's no point in spending more time on this module. It has outlived its usefulness, we should take it behind the barn.

PR #142, opened 2018-11-22
kubevirt_vm: add simplified cloud-init definition

machacekondra:


machacekondra:
@mmazur @pkliczewski @vatsalparekh @adityaramteke

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
```diff
@@ -175,17 +181,29 @@
...
+            image: kubevirt/cirros-registry-disk-demo:latest
+
+- name: Start fedora vm with cloud init
+  kubevirt_vm:
+      state: running
+      wait: true
+      name: myvm
+      namespace: vms
+      memory: 1024M
+      cloud_init_nocloud:
```
pkliczewski: Looks ok, I wonder whether we could allow user only provide things like `username`, `password` and other configuration instead if providing base64 content?

machacekondra: Of course, that's what userData is for :)

```yaml
...
cloud_init_nocloud:
        userData: |
          ssh-authorized-keys:
            - ssh-rsa AAAAB3NzaK8L93bWxnyp test@test.com
          password: fedora
          ... etc
```

machacekondra: I can maybe add example of that rather then base64 example

mmazur:
To be clear – this does not prevent a user from doing it the old way, it just adds a new, simpler way, right?

machacekondra:
Currently, yes.

lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
machacekondra: done

PR #147, opened 2018-11-23
Remove kubevirt_raw module

machacekondra:


machacekondra:
@mmazur @vatsalparekh @pkliczewski @adityaramteke

mmazur:
Yeah, I'm pretty sure we should hold up on this one until we have the other modules in place.

machacekondra:
Which one? I think we should go, the only missing thing is waiting for PVC to be ready, I think that's ok. I added WA for that

machacekondra:
I would really like to do a cleanup of the tests and start testing adding proper examples.

PR #148, opened 2018-11-23
[WIP] Add module to manage VMIRs

machacekondra:
As part of introducing this new module I've extracted common code to `module_utils` class.


machacekondra:
@mmazur @pkliczewski @vatsalparekh @adityaramteke

lib/ansible/modules/clustering/kubevirt/kubevirt_vmir.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+KIND = 'VirtualMachineInstanceReplicaSet'
+VMIR_ARG_SPEC = {
+    'replicas': {'type': 'int', 'required': True},
+    'selector': {'type': 'dict'},
+}
+
+
+class KubeVirtVMIR(KubeVirtRawModule):
```
pkliczewski: Let's change it to `KubeVirtVMIRS`

lib/ansible/modules/clustering/kubevirt/kubevirt_vmir.py
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/python
```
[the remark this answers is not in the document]
machacekondra: Correct, I was confused, by the 's' at the end. It's not recomended to use plural name, but this 's' is part of name, that's what confused me.

lib/ansible/modules/clustering/kubevirt/kubevirt_vmir.py
machacekondra: ok

mmazur:
I think it'd be better to call this module `kubevirt_rs`.

machacekondra:
> I think it'd be better to call this module kubevirt_rs.
Fine by me, any objections by others?

machacekondra:
Renamed.

lib/ansible/utils/module_docs_fragments/kubevirt_vm_options.py
```diff
@@ -0,0 +1,50 @@
...
+              All possible configuration options of network are available in U(https://kubevirt.io/api-reference/master/definitions.html#_v1_network).
+        type: list
+    cloud_init_nocloud:
+        description:
+            - Represents a cloud-init NoCloud user-data source. The NoCloud data will be added
+              as a disk to the virtual machine. A proper cloud-init installation is required inside the guest.
+              More info: U(https://kubevirt.io/api-reference/master/definitions.html#_v1_cloudinitnocloudsource)
+        type: dict
+extends_documentation_fragment:
+    - kubevirt_common_options
```
mmazur: Why? Nothing else uses that file, there's no reason for it to be separate from kubevirt_vm_options.

lib/ansible/utils/module_docs_fragments/kubevirt_vm_options.py
```diff
@@ -0,0 +1,50 @@
...
+# Apache License, Version 2.0
+# (see LICENSE or http://www.apache.org/licenses/LICENSE-2.0)
+
+
+class ModuleDocFragment(object):
+
+    # Standard oVirt documentation fragment
+    DOCUMENTATION = '''
+options:
+    wait:
```
mmazur: This is already defined in kubevirt_common_options.

lib/ansible/module_utils/kubevirt.py
```diff
@@ -0,0 +1,179 @@
...
+
+from openshift import watch
+from openshift.helper.exceptions import KubernetesException
+
+API_VERSION = 'kubevirt.io/v1alpha2'
+
+
+VM_COMMON_ARG_SPEC = {
+    'merge_type': {'type': 'list', 'choices': ['json', 'merge', 'strategic-merge']},
+    'wait': {'type': 'bool', 'default': True},
```
mmazur: I think it's dangerous to have 'wait' default to True. Upstream `k8s` has this default to False.

lib/ansible/module_utils/kubevirt.py
```diff
@@ -0,0 +1,179 @@
...
+    'wait_time': {'type': 'int', 'default': 30},
```
mmazur: … especially with a wait_time this low. On my test setup I regularly hit a timeout because of this. The default `wait_timeout` is 120 in `k8s`.

Also, it's called `wait_timeout` in `k8s` now (added in 2.8). We should use the same naming.

lib/ansible/utils/module_docs_fragments/kubevirt_vm_options.py
machacekondra: I prepared it for the 'presets'.

lib/ansible/utils/module_docs_fragments/kubevirt_vm_options.py
machacekondra: ah, right, thanks

lib/ansible/module_utils/kubevirt.py
machacekondra: It's dangerous in k8s because it's generic and works for everything and they can't wait for everything, those are simplified modules where user would expect to wait by default, because we know what we wait for.

lib/ansible/module_utils/kubevirt.py
machacekondra: Ok, let's increase it. and rename

lib/ansible/module_utils/kubevirt.py
mmazur: I guess the bumped up wait time should be enough for most cases.

PR #150, opened 2018-11-29
Rename registryDisk to containerDisk

imjoey:
According to [[1]], [[2]], `registryDisk` has been renamed to `containerDisk`, which has also been merged into kubevirt master branch.  So this pull request is to be in line with the upstream change and includes the following changes:
1. Rename `registryDisk` to `containerDisk` in XML declarations and documentations
2. Rename `registerdisk` to `containerdisk` in `disk.name` XML definitions and documentations
3. Use the new image `kubevirt/cirros-container-disk-demo:latest` instead of `kubevirt/cirros-registry-disk-demo:latest` in XML definitions and documentations
4. Use the new image `kubevirt/fedora-cloud-container-disk-demo:latest` instead of `kubevirt/fedora-cloud-registry-disk-demo:latest` in XML definitions and documentations
5. Add new `path` support for `containerDisk` declaration in XML definitions and documentations, setting it to `/custom-disk/cirros.img` for image `cirros-container-disk-demo` (see [[3]]) and `/disk/fedora.qcow2` for image `fedora-cloud-container-disk-demo` (see [[4]])

[1]: https://github.com/kubevirt/kubevirt/pull/1730
[2]: https://groups.google.com/d/msgid/kubevirt-dev/CAMuConxp%2B3By5VRvvgRS7jcWWQBMHrTghaE_cBjG2-YGemZw5w%40mail.gmail.com?utm_medium=email&utm_source=footer
[3]: https://github.com/kubevirt/kubevirt/blob/master/images/cirros-container-disk-demo/Dockerfile
[4]: https://github.com/kubevirt/kubevirt/blob/master/images/fedora-cloud-container-disk-demo/Dockerfile 

Signed-off-by: imjoey <majunjiev@gmail.com>

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_rs.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_rs.py
@@ -77,10 +77,11 @@ EXAMPLES = '''
         matchLabels:
             myvmi: myvmi
       disks:
-         - name: registrydisk
+         - name: containerdisk
            volume:
-             registryDisk:
-               image: kubevirt/cirros-registry-disk-demo:latest
+             containerDisk:
+               image: kubevirt/cirros-container-disk-demo:latest
+               path: /custom-disk/cirros.img
            disk:
              bus: virtio
 
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_vm.py
@@ -78,10 +78,11 @@ EXAMPLES = '''
       namespace: vms
       memory: 64M
       disks:
-        - name: registrydisk
+        - name: containerdisk
           volume:
-            registryDisk:
-              image: kubevirt/cirros-registry-disk-demo:latest
+            containerDisk:
+              image: kubevirt/cirros-container-disk-demo:latest
+              path: /custom-disk/cirros.img
           disk:
             bus: virtio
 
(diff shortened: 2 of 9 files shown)


machacekondra:
Thanks for the patch. Let's wait for the new release of kubevirt before merging. I also wonder if there will be API version bump. So maybe we should bump it as well. Let's wait for a while.

imjoey:
@machacekondra Yes, for sure. 😄

mmazur:
This is a good case to discuss our approach to handling api changes. Translating old registryDisk to containerDisk (or vice versa) would likely be easy inside `kubevirt_vm`. Question is – do we wanna?

machacekondra:
Of course we want, or we will break the functionality with new API. We may still support both, but I don't see a reason here because it's still aplha.

imjoey:
IMHO, since neither kubevirt nor this ansible module are production ready yet, for now we may not need to pay much attention to the backward-compatible things and always pick the latest API version is OK. Once the kubevirt is stable, we must take great care about breaking changes.

pkliczewski:
I agree with @machacekondra that it doesn't make sense to keep backward compatibility for the api. ATM the assumption is that API can change. I would wait for new tag in KubeVirt which would contain this change and we can merge this patch.

imjoey:
The new kubevirt tagged version [v0.11.0](https://github.com/kubevirt/kubevirt/releases/tag/v0.11.0) has been released, with noted `API: registryDisk got renamed to containreDisk`. Maybe we could consider to merge this pr. 😄

imjoey:
hi @pkliczewski , I guess that the v0.11.0+ is only required when running testing for our modules. Users could continue to use old playbooks to manage v0.11.0- because the modules do not take part in parsing the configurations within `volume` yaml tags. Please correct me if I'm wrong.

pkliczewski:
@imjoey Yes the modules are not parsing it. I would still keep the version requirement to be consistent.

PR #163, opened 2018-12-11
k8s_service: sync code with upstream PR

mmazur:
With minor workaround for supporting ansible <2.8.
Upstream PR: https://github.com/ansible/ansible/pull/48872

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/k8s_service.py
+++ b/lib/ansible/modules/clustering/kubevirt/k8s_service.py
@@ -2,64 +2,111 @@
 # -*- coding: utf-8 -*-
 
 # Copyright (c) 2018, KubeVirt Team <@kubevirt>
-# Apache License, Version 2.0
-# (see LICENSE or http://www.apache.org/licenses/LICENSE-2.0)
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
 
 ANSIBLE_METADATA = {'metadata_version': '1.1',
                     'status': ['preview'],
                     'supported_by': 'community'}
 
 DOCUMENTATION = '''
----
+
 module: k8s_service
 
 short_description: Manage Services on Kubernetes
-
-description:
-    - Use Openshift Python SDK to manage Services on Kubernetes
 
 version_added: "2.8"
 
 author: KubeVirt Team (@kubevirt)
 
-options:
-    name:
-        description:
-            - Use to specify a Service object name.
-        required: true
-        type: str
-    namespace:
-        description:
-            - Use to specify a Service object namespace.
-        required: true
-        type: str
-    type:
-        description:
-            - Specifies the type of Service to create.
-            - U(https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types)
-        choices:
-            - NodePort
-            - ClusterIP
-            - LoadBalancer
-            - ExternalName
-        default: ClusterIP
-        required: true
-    ports:
-        description:
-            - A list of ports to expose.
-            - U(https://kubernetes.io/docs/concepts/services-networking/service/#multi-port-services)
-        required: true
-        type: list
-    selector:
-        description:
-            - Label selectors identify objects this Service should apply to.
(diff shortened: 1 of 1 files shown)


lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -2,64 +2,113 @@
...
+  name:
+    description:
+      - Use to specify a Service object name. If I(resource definition) is provided, the I(metadata.name)
+        value from the I(resource_definition) will override this option.
+    required: true
+    type: str
+  namespace:
+    description:
+      - Use to specify a Service object namespace. If I(resource definition) is provided, the
+        I(metadata.namespace) value from the I(resource_definition) will override this option.
```
machacekondra: Isn't it the other way around?

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -2,64 +2,113 @@
...
+  - Use Openshift Python SDK to manage Services on Kubernetes
 
 extends_documentation_fragment:
   - k8s_auth_options
-  - k8s_resource_options
+
+options:
+  resource_definition:
+    description:
+    - A partial YAML definition of the Service object being created/updated. Here you can define Kubernetes
```
machacekondra: Any reason to not use doc from `k8s_state_options`?

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Good catch, you're right, I didn't correct this from older code.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
mmazur: Three:
1. The description isn't 100% fitting.
2. If you need `src`, you should just use `k8s` module.
3. By default `resource_definition` supports both a single object and a list of objects. I'm enforcing only the former.

In general the approach is that this gets merged without 100% code reuse, but once another k8s_something module gets added, that'll be the time to extract some more common code and generalize stuff. The current "common" code would be misleading and not 100% correct for this particular module.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -162,30 +231,37 @@ def merge_dicts(x, y):
...
 
+        api_version = 'v1'
         selector = self.params.get('selector')
         service_type = self.params.get('type')
         ports = self.params.get('ports')
 
-        definition = virtdict()
+        definition = defaultdict(defaultdict)
+
+        definition['kind'] = 'Service'
```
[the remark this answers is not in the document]
mmazur: This code only doesn't make sense due to the fact this module has a `kind` parameter. That parameter does not exist upstream and is only here as a workaround. The goal of this PR is to have our code match upstream as much as possible, which is why I didn't correct working code even if it didn't make 100% sense.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
```diff
@@ -162,30 +231,37 @@ def merge_dicts(x, y):
...
-        argument_spec.update(copy.deepcopy(AUTH_ARG_SPEC))
+        argument_spec = copy.deepcopy(AUTH_ARG_SPEC)
         argument_spec.update(SERVICE_ARG_SPEC)
         return argument_spec
 
     def execute_module(self):
         """ Module execution """
         self.client = self.get_api_client()
 
+        api_version = 'v1'
```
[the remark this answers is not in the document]
mmazur: Probably. But the reviewer asked me to remove `api_version` arg from the module, so I did it in the least lines of code possible, which is why this line exists now. I figured that upstream reviewers have very little time and it's not in my interest to mess around with code that has already been generally approved.

mmazur:
Ok, pushed a version with some things corrected.

lib/ansible/modules/clustering/kubevirt/k8s_service.py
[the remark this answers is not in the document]
machacekondra: Sure, I just wondered why you have it twice, once here and on line 205, but not important, as it's just WA.

PR #166, opened 2018-12-13
Add kubevirt_cdi_upload module

mmazur:


tests/playbooks/kubevirt_pvc_img_upload.yml
```diff
@@ -0,0 +1,46 @@
...
+       name: pvc-vm1
+       inline:
+         apiVersion: v1
+         kind: PersistentVolumeClaim
+         metadata:
+           name: pvc-vm1
+           labels:
+             app: containerized-data-importer
+           annotations:
+             cdi.kubevirt.io/storage.upload.target: ""
```
pkliczewski: Should that be target PVC?

tests/playbooks/kubevirt_pvc_img_upload.yml
```diff
@@ -0,0 +1,46 @@
...
+      register: pvc
+      until: "pvc.resources[0].status.phase == 'Bound'"
+      delay: 10
+      retries: 30
+
+    - name: Upload local image to pvc
+      kubevirt_cdi_upload:
+        pvc_namespace: default
+        pvc_name: pvc-vm1
+        upload_host: https://127.0.0.1:55555
```
pkliczewski: Would be good to add documentation about how to set it up so people using (verifying) it would know how to do it.

tests/playbooks/kubevirt_pvc_img_upload.yml
mmazur: That's just the weird way CDI makes you annotate a PVC ready for image upload. In `kubevirt_pvc` I was planning on masking it with something like `upload: bool`. https://github.com/kubevirt/containerized-data-importer/blob/master/doc/upload.md#create-a-persistentvolumeclaim

tests/playbooks/kubevirt_pvc_img_upload.yml
mmazur: I was planning on expanding the CDI doc on this and then linking to that. Also there's a sneaky way for the module to circumvent the requirement for the cdi-uploadproxy to have a working Service+Ingress combo (which is what would end up as the `upload_host`) configured and instead do some kubectl port-forward magic.

Once I have the base functionality in place, I'll write a short description of that, so we can figure out whether we want to spend time on adding it.

mmazur:
@pkliczewski @machacekondra this works and is reviewable. What's left is return values/arg docs/etc., which I'll finish up now.

pkliczewski:
@mmazur I still see `WIP` in the title and failing CI.

mmazur:
Aaand done.

pkliczewski:
@machacekondra @vatsalparekh please review

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+- name: Upload local image to pvc-vm1
+  kubevirt_cdi_upload:
+    pvc_namespace: default
+    pvc_name: pvc-vm1
+    upload_host: https://localhost:8443
+    upload_host_verify_ssl: false
+    path: /tmp/cirros-0.4.0-x86_64-disk.img
+'''
+
+RETURN = '''# '''
```
machacekondra: So you think we shouldn't return anything, only `change=true/false`?

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+        imgfile = open(path, 'rb')
+
+        resource = self.find_resource('UploadTokenRequest', api_version, fail=True)
+        definition = self.set_defaults(resource, definition)
+        result = self.perform_action(resource, definition)
+
+        headers = {'Authorization': "Bearer {}".format(result['result']['status']['token'])}
+        files = {'file': imgfile}
+        url = "{}/v1alpha1/upload".format(upload_host)
+        requests.post(url, files=files, headers=headers, verify=upload_host_verify_ssl)
```
machacekondra: Should we have wait here true/false?

machacekondra: We will be here uploading big files, do we know if it's performant enough?

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+      - UploadTokenRequest
+  api_version:
+    description:
+      - Required for compatibility with ansible <2.8. Can't be modified.
+    default: upload.cdi.kubevirt.io/v1alpha1
+    choices:
+      - upload.cdi.kubevirt.io/v1alpha1
+  src:
+    description:
+      - Required for compatibility with ansible <2.8. Don't use.
```
machacekondra: Why it's required for that compatibility?

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+        resource = self.find_resource('UploadTokenRequest', api_version, fail=True)
+        definition = self.set_defaults(resource, definition)
+        result = self.perform_action(resource, definition)
+
+        headers = {'Authorization': "Bearer {}".format(result['result']['status']['token'])}
+        files = {'file': imgfile}
+        url = "{}/v1alpha1/upload".format(upload_host)
+        requests.post(url, files=files, headers=headers, verify=upload_host_verify_ssl)
+
+        self.exit_json(changed=True)
```
machacekondra: So this module isn't idempotent, we should document it if we really want it.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+
+module: kubevirt_cdi_upload
+
+short_description: Upload local VM images to CDI Upload Proxy.
```
machacekondra: I Would add some note/link or something how to deploy CDI

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+
+module: kubevirt_cdi_upload
```
machacekondra: I think we should consider adding just single module for all CDI functionality.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+      - Required for compatibility with ansible <2.8. Don't use.
+  namespace:
+    description:
+      - Required for compatibility with ansible <2.8. Don't use.
+
+
+requirements:
+  - python >= 2.7
+  - openshift >= 0.6.2
+  - requests >= 2.0.0
```
machacekondra: Requests is required by kubernetes so I think we don't need to list it here.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+    description:
+      - URL containing the host and port on which the CDI Upload Proxy is available.
+  upload_host_verify_ssl:
+    description:
+      - Whether or not to verify the CDI Upload Proxy's SSL certificates.
+    default: true
+    type: bool
+  path:
+    description:
+      - Path of local image file to transfer.
```
machacekondra: Is this qcow/raw, or what formats are supported?

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+  pvc_namespace:
+    description:
+      - Use to specify the namespace of the target PersistentVolumeClaim.
+    required: true
+  upload_host:
+    description:
+      - URL containing the host and port on which the CDI Upload Proxy is available.
+  upload_host_verify_ssl:
+    description:
+      - Whether or not to verify the CDI Upload Proxy's SSL certificates.
```
machacekondra: So if we add true, system CA is used to verify, or which one?

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+        # Let's check the file's there before we do anything else
+        imgfile = open(path, 'rb')
+
+        resource = self.find_resource('UploadTokenRequest', api_version, fail=True)
+        definition = self.set_defaults(resource, definition)
+        result = self.perform_action(resource, definition)
+
+        headers = {'Authorization': "Bearer {}".format(result['result']['status']['token'])}
+        files = {'file': imgfile}
+        url = "{}/v1alpha1/upload".format(upload_host)
```
machacekondra: I think this should be constatnt.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+        def_meta['name'] = pvc_name
+        def_meta['namespace'] = pvc_namespace
+
+        def_spec = definition['spec']
+        def_spec['pvcName'] = pvc_name
+
+        # Let's check the file's there before we do anything else
+        imgfile = open(path, 'rb')
+
+        resource = self.find_resource('UploadTokenRequest', api_version, fail=True)
```
machacekondra: I think this should be constant.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
```diff
@@ -0,0 +1,205 @@
...
+        definition = self.set_defaults(resource, definition)
+        result = self.perform_action(resource, definition)
+
+        headers = {'Authorization': "Bearer {}".format(result['result']['status']['token'])}
```
machacekondra: Should we revoke the token, when done?

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: Initially I assumed we'd be returning something interesting. But it's a file upload module, there's really no useful data that it can return for later processing.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: Unless you mean something else, it's not possible to have a non–wait upload code. By definition it needs to be running while the file is being uploaded and then terminates once everything gets uploaded.

mmazur: It seems to be the most popular http library for python and is being used by some other ansible modules as well (not for file uploading though). I have no way to tell whether there's something better out there, but also have good reason to believe that this is a solid choice. If actual issues are found, switching it to something else will not be hard.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: The base Kubernetes classes were created with only `k8s` module in mind. This has been corrected to a large degree with 2.8 (partially with my `k8s_service` patches), so it's much better on the devel branch of ansible, but in 2.7 it's very noticeable. If you don't define all those parameters, the module will just fail with 'key XYZ not found'.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: The token only supports the Create operation. You create it when you need it to upload something and then forget it ever existed.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: The most popular non–idempotent module would likely be the core [`command`](https://docs.ansible.com/ansible/latest/modules/command_module.html). I've checked and there's no mention of that anywhere. Seems like people are just expected to figure out on their own which modules are safely rerunnable. (Unless there's some standard wording I haven't found.)

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: I believe CDI is included as standard in CNV and in kubevirt's dev env? @pkliczewski 

I did add a link to exposing the upload proxy itself in the description of `upload_host` parameter though.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: I thought so initially as well, but later figured it'd make `kubevirt_pvc` more complicated that way and a bit harder to use, as it'd need to internally do many different things at once. Splitting the functionality, I get a clean idempotent `kubevirt_pvc` and a small dedicated upload module, which works fine without `kubevirt_pvc`, simply because you can create a CDI–ready PVC with just `k8s` module, but you can't actually upload data without dedicated code.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: All the other modules that use _requests_ explicitly list them as required. Guess that's module policy.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: Dunno. I guess it'd make most sense to reference CDI's documentation on this, but CDI does not document this in an explicit way. Their [example](https://github.com/kubevirt/containerized-data-importer/blob/master/doc/upload.md#upload-an-image) suggests qcow2 is supported, but that's all I know.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: Yes, it seems to use standard system certs (whatever those are currently). Interestingly enough the [underlying library](http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification) also supports specifying a path to a CA bundle in addition to `true` and `false`. I wonder if in this particular case this isn't something we should also support.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
machacekondra: Ok, when `kubevirt_pvc` is done, we can re-thing the integration.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
machacekondra: Maybe would be good to document it.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
machacekondra: Ok, but then this module don't have any advantage of just running actually the command module, with single `kubectl && curl`. But if you think we can't do it any better I am OK.

mmazur: Well, the primary purpose for this repo lately has been to hide underlying k8s complexity behind more familiar–looking ansible modules. This one fits right into that.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: I've moved this into a common var.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: I've moved this into a common var.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: If this gets somehow submitted upstream, then yes, we'll need to clean it up. But that will be a whole different process, nobody's gonna let us not remove those args. :)

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: Yeah, that's potentially a good idea. Though I'd need some clarification on what's actually supported first and that'll take a bit of time.

lib/ansible/modules/clustering/kubevirt/kubevirt_cdi_upload.py
mmazur: That would force me to describe details of system–wide CA implementation on every platform supported by ansible. This module's docs are not the place for that. The general rule I've noticed is that every time there's a "verify ssl certs" on/off in a tool, the implication is always "… against whatever is the standard cert storage on your system". You only document this stuff if you allow for something more (like `ca_path`).
(This is at least true for other ansible modules, with the parameter typically called `verify_ssl`.)

machacekondra: Yeah so just write, 'if not specified uses system CA ', that's good enough.you don't have to describe how to add cert to system CA store. But up to you.

mmazur:
Ok, I think this should be merged now. Things for a future TODO:
1. `upload_host_ca_path` to specify alternative CA trust store for CDI proxy connection. I could just add it, but this depends on proper support from the underlying library and if there's a chance we'll need to switch to curl down the line for handling http(s), then this would need more research then I have time for (I'm only in today and tomorrow, then gone till January.)
2. Local image file format verification. But that would require CDI documenting/committing to this stuff which would again take some time.

mmazur:
Just noticed nobody merged this. :)

PR #168, opened 2019-01-07
Bump min openshift rest client lib ver

mmazur:
Needed for CDI upload to work correctly.

@machacekondra or @pkliczewski plz merge, so QA can check cdi uploading.


mmazur:
Good point, done.

Issue #133, opened 2018-11-08
Align templates with guidelines


pkliczewski:
There are template [guidelines](https://github.com/kubevirt/common-templates/blob/master/templates/README.md) which we do not follow so we need to align the code.

pkliczewski:
@machacekondra I think that this was fixed already. Are you ok to close it?

PR #169, opened 2019-01-18
Add template module

machacekondra:
This PR add new kubevirt_template module.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
+++ b/lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
@@ -1,0 +1,387 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2018, KubeVirt Team <@kubevirt>
+# Apache License, Version 2.0
+# (see LICENSE or http://www.apache.org/licenses/LICENSE-2.0)
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+---
+module: kubevirt_template
+
+short_description: Manage KubeVirt templates
+
+description:
+    - Use Openshift Python SDK to manage the state of KubeVirt templates.
+
+version_added: "2.8"
+
+author: KubeVirt Team (@kubevirt)
+
+options:
+    state:
+        description:
+            - Set the to be either I(present) or I(absent).
+        default: "present"
+        choices:
+            - present
+            - absent
+        type: str
+    name:
+        description:
+            - Name of the template.
+        required: true
+        type: str
+    namespace:
+        description:
+            - Namespace where the template exists.
+        required: true
+        type: str
+    objects:
+        description:
+            - List of any valid API objects, such as a I(DeploymentConfig), I(Service), etc. The object
+              will be created exactly as defined here, with any parameter values substituted in prior to creation.
+              The definition of these objects can reference parameters defined earlier.
+            - As part of the the list user can pass also I(VirtualMachine) kind. When passing I(VirtualMachine)
+              user must use Ansible strucutre of the paramters not the Kubernetes API structure. For more information
+              please take a look at M(kubevirt_vm) module and at EXAMPLES section, where you can see example.
+        type: list
+    display_name:
+        description:
+            - "A brief, user-friendly name, which can be employed by user interfaces."
+        type: str
+    description:
+        description:
+            - "A description of the template. Include enough detail that the user will understand what is being deployed
+               and any caveats they need to know before deploying. It should also provide links to additional information,
+               such as a README file."
+        type: str
+    long_description:
+        description:
+            - "Additional template description. This may be displayed by the service catalog, for example."
+        type: str
+    provider_display_name:
+        description:
+            - "The name of the person or organization providing the template."
+        type: str
+    documentation_url:
+        description:
+            - "A URL referencing further documentation for the template."
+        type: str
+    support_url:
+        description:
+            - "A URL where support can be obtained for the template."
+        type: str
+    version:
+        description:
+            - Template structure version.
+            - This is parameter can be used only when kubevirt addon is installed on your openshift cluster.
+        type: str
+    editable:
+        description:
+            - "Extension for hinting at which elements should be considered editable.
+               List of jsonpath selectors. The jsonpath root is the objects: element of the template."
+            - This is parameter can be used only when kubevirt addon is installed on your openshift cluster.
+        type: list
+    default_disk:
+        description:
+            - "The goal of default disk is to define what kind of disk is supported by the OS mainly in
+               terms of bus (ide, scsi, sata, virtio, ...)"
+            - The C(default_disk) parameter define configuration overlay for disks that will be applied on top of disks
+              during virtual machine creation to define global compatibility and/or performance defaults defined here.
+            - This is parameter can be used only when kubevirt addon is installed on your openshift cluster.
+        type: dict
+    default_volume:
+        description:
+            - "The goal of default volume is to be able to configure mostly performance parameters like
+               caches if those are exposed by the underlying volume implementation."
+            - The C(default_volume) parameter define configuration overlay for volumes that will be applied on top of volumes
+              during virtual machine creation to define global compatibility and/or performance defaults defined here.
+            - This is parameter can be used only when kubevirt addon is installed on your openshift cluster.
+        type: dict
+    default_nic:
+        description:
+            - "The goal of default network is similar to I(default_disk) and should be used as a template
+               to ensure OS compatibility and performance."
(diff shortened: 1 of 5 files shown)


lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
```diff
@@ -0,0 +1,271 @@
...
+        type: str
+    objects:
+        description:
+            - List of any valid API objects, such as a I(DeploymentConfig), I(Service), etc. The object will be created exactly as defined here, with any parameter values substituted in prior to creation. The definition of these objects can reference parameters defined earlier.
+        type: list
+    vms:
+        description:
+            - List of virtual machine object to be created from template. Please refer to module I(kubevirt_vm) to see list of all supported parameters of virtual machine object. The object will be created exactly as defined here, with any parameter values substituted in prior to creation. The definition of these objects can reference parameters defined earlier.
+        type: list
+    objects:
```
mmazur: This is duplicated, there's already an `objects` a few lines above.

mmazur:
Code looks ok (except for flake8 complaining), though since I don't really know how Templates are supposed to be used, I can't comment on the design.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
```diff
@@ -0,0 +1,271 @@
...
+    namespace:
+        description:
+            - Namespace where the template exists.
+        required: true
+        type: str
+    objects:
+        description:
+            - List of any valid API objects, such as a I(DeploymentConfig), I(Service), etc. The object will be created exactly as defined here, with any parameter values substituted in prior to creation. The definition of these objects can reference parameters defined earlier.
+        type: list
+    vms:
```
pkliczewski: I wonder why we need to differentiate `vms`? Why not to keep everything in objects?

machacekondra: So we can have it flat. It can be in objects, but then the VM spec won't be falt, but nested same as if you use API directly.

pkliczewski: Can we iterate over and find `known` objects and check simplified? We could use `handlers` for each object type and start with `vms` and add other kinds later.

pkliczewski:
The code looks good. Please make sure that we are align with https://github.com/kubevirt/common-templates

machacekondra:
It is, I've added all `cnv.io` related annotations as module parameters.

pkliczewski:
@MarSik Please review. In this PR we are simplifying template management module and as part of this work we want to align the code with changes that your team introduced.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
```diff
@@ -0,0 +1,349 @@
...
+            - "Template structure version."
+        type: str
+    editable:
+        description:
+            - "Extension for hinting at which elements should be considered editable.
+               List of jsonpath selectors. The jsonpath root is the objects: element of the template."
+        type: list
+    default_disk:
+        description:
+            - "The goal of default network is similar to I(default_disk) and should be used as a template
```
MarSik: You are describing default network in section for default_disk.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
```diff
@@ -0,0 +1,349 @@
...
+        required: true
+        type: str
+    namespace:
+        description:
+            - Namespace where the template exists.
+        required: true
+        type: str
+    objects:
+        description:
+            - List of any valid API objects, such as a I(DeploymentConfig), I(Service), etc. The object
```
MarSik: What kind of template management is this supposed to be? I see you specifically mention DepoymentConfig, Service... all of which are pure kubernetes concepts. A little bellow you added default_disk, nic, volume parameters that are only relevant for VM templates and actually do not make any sense (as we ship them within the Template object).

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
```diff
@@ -0,0 +1,349 @@
...
+    version:
+        description:
+            - "Template structure version."
+        type: str
+    editable:
+        description:
+            - "Extension for hinting at which elements should be considered editable.
+               List of jsonpath selectors. The jsonpath root is the objects: element of the template."
+        type: list
+    default_disk:
```
MarSik: An example of how do you expect the default_* parameters to be used would be nice. Because I have no idea what will those do here.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
machacekondra: Well, it should support templates of both Openshift/k8s objects and kubevirt, so the description is general, I will improve.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
machacekondra: I've added an example.

pkliczewski:
@MarSik Please review this PR again

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
```diff
@@ -0,0 +1,374 @@
...
+        type: str
+    support_url:
+        description:
+            - "A URL where support can be obtained for the template."
+        type: str
+    version:
+        description:
+            - "Template structure version."
+        type: str
+    editable:
```
MarSik: I still think this will be confusing as the user does not see the result yaml while writing this.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
MarSik: Maybe a short explanation about the result would be nice too. Something like: The default_ parameters define configuration overlays for disks, volumes, ... that will be applied on top of disks, volumes.. during VM creation to define global compatibility and/or performance defaults defined here.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
```diff
@@ -0,0 +1,374 @@
...
+    display_name: Generic cirros template
+    description: Basic cirros template
+    long_description: Verbose description of cirros template
+    provider_display_name: Just Be Cool, Inc.
+    documentation_url: http://theverycoolcompany.com
+    support_url: http://support.theverycoolcompany.com
+    icon_class: icon-linux
+    namespace: templates
+    default_disk:
+      - name: default-disk
```
MarSik: Is the name really necessary? It is not used anywhere else and it really is an implementation detail.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
```diff
@@ -0,0 +1,374 @@
...
+    default_network:
+      - name: default-network
+        resource:
+          resourceName: bridge.network.kubevirt.io/cnvmgmt
+    default_volume:
+      - name: default-volume
+        containerDisk:
+          image: kubevirt/cirros-container-disk-demo:latest
+    objects:
+      - name: ${NAME}
```
MarSik: This is not a valid VM object in kubevirt form so the description above is inaccurate. You should somehow describe the valid simplified structures (by pointing to the other modules or something)

pkliczewski:
This PR needs to Fix: https://github.com/kubevirt/ansible-kubevirt-modules/issues/133

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
machacekondra: I've removed it.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
machacekondra: Thanks, I've added documentation, where user should look for structure.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
machacekondra: Thanks, added.

lib/ansible/modules/clustering/kubevirt/kubevirt_template.py
machacekondra: Yes, I've added comment that we need to take a look at this. The problem is that user can define any object, so for some it could be JSON path and for some Ansible path, so that might actually be confusing even more, because now it's only JSON.

machacekondra:
@MarSik Anything else you would like to have fixed/changed?

machacekondra:
@MarSik Can you please take a last look, thank you.

machacekondra:
Thanks, I've opened issue, so we don't forget: https://github.com/kubevirt/ansible-kubevirt-modules/issues/187

PR #173, opened 2019-01-25
Add api version detection logic

### LINKED ISSUE #77: Support multiple KubeVirt API versions
Currently, KubeVirt's API is still at pre-v1 therefore the Python client doesn't support anything else.

The modules reference `V1*` objects everywhere, but eventually other versions will have to be supported as well.

Ideally, [this RFE](https://github.com/kubevirt/client-python/issues/19) would be implemented and it'll hide the Python client implementation details, easing the modules code. If that RFE doesn't get implemented some abstraction will be needed in the modules to handle different API versions.

mmazur:
Use underlying library to dynamically discover supported api version for the k8s objects we operate on. Make sure to do something user–friendly in all possible cases, including "cluster too new to handle" and "cluster really old, but we can manage".

Fixes #77

### DIFF
--- a/lib/ansible/module_utils/kubevirt.py
+++ b/lib/ansible/module_utils/kubevirt.py
@@ -6,13 +6,17 @@
 # (see LICENSE or http://www.apache.org/licenses/LICENSE-2.0)
 
 from collections import defaultdict
+from distutils.version import Version
 
 from ansible.module_utils.k8s.raw import KubernetesRawModule
 
 from openshift import watch
 from openshift.helper.exceptions import KubernetesException
 
-API_VERSION = 'kubevirt.io/v1alpha3'
+import re
+
+MAX_SUPPORTED_API_VERSION = 'v1alpha3'
+API_GROUP = 'kubevirt.io'
 
 
 VM_COMMON_ARG_SPEC = {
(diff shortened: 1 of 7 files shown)


pkliczewski:
There are bunch of old versions in our playbook [directory](https://github.com/kubevirt/ansible-kubevirt-modules/tree/master/tests/playbooks)

mmazur:
Ok, the api version detection logic is working exactly as we've discussed it should. Future patches are free to contain version–specific logic based on `resource.api_version`.

lib/ansible/module_utils/kubevirt.py
```diff
@@ -136,6 +137,17 @@ def _define_disks(self, disks, template_spec):
                 spec_volumes.append(volume)
             template_spec['volumes'] = spec_volumes
 
+    def find_supported_resource(self, kind):
+        results = self.client.resources.search(kind=kind, group=API_GROUP)
+        if not results:
+            self.fail('Failed to find resource {0} in {1}'.format(kind, API_GROUP))
+        sr = sorted(results, key=lambda r: LooseVersion(r.api_version), reverse=True)
+        for r in sr:
+            if LooseVersion(r.api_version) <= LooseVersion(MAX_SUPPORTED_API_VERSION):
```
pkliczewski: How this would work if the current version was `v1beta1`? Would this code fail as it should?

pkliczewski:
It seems like travis is failing due to not related test. It would be great to fix it in different PR.

lib/ansible/module_utils/kubevirt.py
mmazur: LooseVersion is also used internally in ansible k8s code, since it's supposed to be smart. But seems it's not smart enough for our case. I'll have to dig a bit more.

```
>>> from distutils.version import LooseVersion as lv
>>> lv('v1beta1') > lv('v1alpha3')
True
>>> lv('v1') > lv('v1alpha3')
False
```

mmazur:
CI issues were related. I've updated the patch, though I suspect tests might fail now.

lib/ansible/module_utils/kubevirt.py
machacekondra: I just wonder, why for us is lower API version a working version if even older can't work(maybe)? Any reason for that?

mmazur: I've implemented our own KubeAPIVersion() that correctly handles all the version logic.

@machacekondra older version can work. We can support say v1alpha1, v1alpha2 and v1alpha3 or just v1alpha3. Whenever we bump MAX_SUPPORTED_API_VERSION, we have to decide whether nobody's gonna use the previous version and we don't care about supporting it, or somebody might use the previous version (say, with a bump from v1 to v2) and we do need to support it (in addition to the new version).

In the latter case, the PR that bumps MAX_SUPPORTED_API_VERSION v1 -> v2 needs to also contain logic that goes `if resource.api_version == 'v1' in places where that's required.

In other words – we don't really use the capabilities of this patch **yet**. But now we'll have the option to.

machacekondra: Well, it *could* work, but maybe it won't, and so newer version *could* work but maybe it won't. It's just guessing. I would forbid to use older versions same as we forbid to use newer versions.

mmazur: Yes, but the difference is in the process. When bumping MAX_SUPPORTED_API_VERSION we control whether we care if the older versions work or not (by adding compatibility code; or not). If the code just picks an available api version and attempts to do something, we lose that control (since old code will work some of the time on new clusters and the users isn't forced to update the modules until they get a cryptic error message on some of the playbooks).

If it turns out we'd want more control, then changing MAX_ to an actual list of supported api versions will be a trivial change.

mmazur: Hmm, and when we get to the more stable api versions, I'm pretty sure we'll go more conservative and switch to using a list.

mmazur:
Ok, tests pass, version comparison fixed. Unless we want to switch implementation to a list of acceptable api version instead of a max_supported_ver, then this patch is ready for merging.

lib/ansible/module_utils/kubevirt.py
machacekondra: I am fine with that, but for example alphav2, alphav1 Kubevirt won't work with this and user will be getting strange errors, same as they used to.

machacekondra: We shouldn't care to much about those alpha versions IMHO. So I am fine with whatever we do here, just pointing out that some cases might not work.

PR #174, opened 2019-01-28
Add self.warnings to k8s–derived classes

mmazur:
KubernetesRawModule()–derived modules assume `self.warnings` exists when handling exceptions. This is true for the `k8s` module. It's not true for our modules. I will send a general fix for this to upstream, but for the time being we need this list created in our own code as well.

Without this if there's an exception ansible returns tens of lines of multi–level exception trace ending with something like `"msg": "'KubeVirtVM' object has no attribute 'warnings'"` and a user needs to parse all of it to figure out what the actual problem really is.

### DIFF
--- a/lib/ansible/module_utils/kubevirt.py
+++ b/lib/ansible/module_utils/kubevirt.py
@@ -139,6 +139,7 @@ class KubeVirtRawModule(KubernetesRawModule):
     def execute_crud(self, kind, definition, template):
         """ Module execution """
         self.client = self.get_api_client()
+        self.warnings = []
 
         disks = self.params.get('disks', [])
         memory = self.params.get('memory')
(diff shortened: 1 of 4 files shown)


machacekondra:
This is for Ansible 2.8 or devel branch?

mmazur:
Neither. Upstream I will fix this in the base k8s code, so we don't have to do this in the modules. But for the time being a fix for our modules (in our repo) is this patch.

mmazur:
Upstream PR for reference: https://github.com/ansible/ansible/pull/51386

machacekondra:
But when it was introduced? In 2.8 or now in devel, because it used to work

machacekondra:
I mean 2.7, 2.8 is not yet out so it's devel

mmazur:
Ha, good point, I thought this was introduced in 2.7, but nope, 2.8 only, so if they take my upstream patch, then this PR is not needed at all.

mmazur:
What I mean is – please don't merge this and wait for upstream to respond. If upstream merges my fix, then I will drop this PR.

mmazur:
Upstream merged a fix, closing this PR.

PR #175, opened 2019-01-28
Add k8s_auth from upstream PR

### LINKED ISSUE #151: parameter: username in k8s module does not work
/kind bug

**What happened**:
```
---
- name: Create a VM using k8s module
  hosts: localhost
  tasks:
    - name: Create a VM using yaml file
      k8s:
        username: test_user
        password: 12345
        state: present
        src: /home/aramteke/cnv-qe/vm-yaml-examples/vm-cirros.yaml
        namespace: default

```
The VM is get created by the default user that you are logged in and not with the mentioned user `test_user`

**What you expected to happen**:
As mentioned in the above YAML file. The creation of VM must be done by `test_user` user.

mmazur:
Upstream PR: https://github.com/ansible/ansible/pull/50807

**What this PR does / why we need it**: makes ansible/k8s usable with openshift auth

Fixes #151 

**Special notes for your reviewer**: don't tell the other reviewers, but you're my favorite reviewer.

### DIFF
--- a/lib/ansible/modules/clustering/kubevirt/k8s_auth.py
+++ b/lib/ansible/modules/clustering/kubevirt/k8s_auth.py
@@ -1,0 +1,342 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright (c) 2018, KubeVirt Team <@kubevirt>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+ANSIBLE_METADATA = {'metadata_version': '1.1',
+                    'status': ['preview'],
+                    'supported_by': 'community'}
+
+DOCUMENTATION = '''
+
+module: k8s_auth
+
+short_description: Authenticate to Kubernetes clusters which require an explicit login step
+
+version_added: "2.8"
+
+author: KubeVirt Team (@kubevirt)
+
+description:
+  - "This module handles authenticating to Kubernetes clusters requiring I(explicit) authentication procedures,
+    meaning ones where a client logs in (obtains an authentication token), performs API operations using said
+    token and then logs out (revokes the token). An example of a Kubernetes distribution requiring this module
+    is OpenShift."
+  - "On the other hand a popular configuration for username+password authentication is one utilizing HTTP Basic
+    Auth, which does not involve any additional login/logout steps (instead login credentials can be attached
+    to each and every API call performed) and as such is handled directly by the C(k8s) module (and other
+    resource–specific modules) by utilizing the C(host), C(username) and C(password) parameters. Please
+    consult your preferred module's documentation for more details."
+
+options:
+  state:
+    description:
+    - If set to I(present) connect to the API server using the URL specified in C(host) and attempt to log in.
+    - If set to I(absent) attempt to log out by revoking the authentication token specified in C(api_key).
+    default: present
+    choices:
+    - present
+    - absent
+  host:
+    description:
+    - Provide a URL for accessing the API server.
+    required: true
+  username:
+    description:
+    - Provide a username for authenticating with the API server.
+  password:
+    description:
+    - Provide a password for authenticating with the API server.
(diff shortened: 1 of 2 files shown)


mmazur:
Note: the examples with the yaml anchors aren't all that pretty. There's a better/proper way and we don't support it yet, but there's a card for that and it's waiting for our next planning session.

mmazur:
CI now passes, upstream code got a review and a _shipit_, so @machacekondra @pkliczewski please merge.

lib/ansible/modules/clustering/kubevirt/k8s_auth.py
```diff
@@ -0,0 +1,338 @@
...
+  username:
+    description:
+    - Provide a username for authenticating with the API server.
+  password:
+    description:
+    - Provide a password for authenticating with the API server.
+  ssl_ca_cert:
+    description:
+    - "Path to a CA certificate file used to verify connection to the API server. The full certificate chain
+      must be provided to avoid certificate validation errors."
```
pkliczewski: Once `verify_ssl` is set to `false` this path is not needed. I think we should add it to the description.

tests/playbooks/k8s_auth.yml
```diff
@@ -0,0 +1,46 @@
...
+  - block:
+    # It's good practice to store login credentials in a secure vault and not
+    # directly in playbooks.
+    #- include_vars: k8s_passwords.yml
+
+    - name: Log in (obtain auth token)
+      k8s_auth:
+        host: https://192.168.42.55:8443
+#       ssl_ca_cert: ca.pem
+        verify_ssl: no
```
pkliczewski: This one is confusing. We default `verify_ssl` to `True`. Why we are using `no`?

mmazur: Because we want this to fail by default and if the operator wants it to work, he needs to take the time to make it work. Either by providing the cert chain or disabling it. Either way the point is that the default behavior should be secure to the extent possible.

lib/ansible/modules/clustering/kubevirt/k8s_auth.py
mmazur: Yeah, there's a note to that effect in upstream k8s module docs. I've copied the note and pushed an updated version.

lib/ansible/modules/clustering/kubevirt/k8s_auth.py
```diff
@@ -0,0 +1,342 @@
...
+    description:
+    - If set to I(present) connect to the API server using the URL specified in C(host) and attempt to log in.
+    - If set to I(absent) attempt to log out by revoking the authentication token specified in C(api_key).
+    default: present
+    choices:
+    - present
+    - absent
+  host:
+    description:
+    - Provide a URL for accessing the API server.
```
machacekondra: having `url` in `host` parameter might be confusing

lib/ansible/modules/clustering/kubevirt/k8s_auth.py
```diff
@@ -0,0 +1,342 @@
...
+
+EXAMPLES = '''
+- block:
+  # It's good practice to store login credentials in a secure vault and not
+  # directly in playbooks.
+  - include_vars: k8s_passwords.yml
+
+  - name: Log in (obtain access token)
+    k8s_auth:
+      host: https://k8s.example.com/
```
machacekondra: If in the host you have just http://fqdn/, maybe worth to support passing only fqdn without https:?

mmazur: I don't get the freedom to do that. Any parameters shared between the k8s_* modules should have the exact same use and if I want to do something different, I'd likely need to provide a patch that modifies all of the modules' behavior.

lib/ansible/modules/clustering/kubevirt/k8s_auth.py
mmazur: True, but the confusion is more from the fact that the parameter shouldn't be named `host`. However the parameter *is* a url, a base url for the api endpoint to be precise. Out of scope for this PR, but I think upstream should change some of the parameter names to something more relevant and keep current names only as aliases.

PR #180, opened 2019-02-11
Fixed e2e playbook

vatsalparekh:
Fixed e2e playbook

**Release note**:

```release-note
Fixed e2e playbook with updated modules
```

### DIFF
--- a/tests/playbooks/e2e.yaml
+++ b/tests/playbooks/e2e.yaml
@@ -41,52 +41,18 @@
       retries: 18
 
     - name: Create a VM with the CDI PVC that was created
-      k8s:
+      kubevirt_vm:
        state: present
-       name: '{{ vm_name }}'
+       name: "{{ vm_name }}"
        namespace: default
-       inline:
-         apiVersion: kubevirt.io/v1alpha2
-         kind: VirtualMachine
-         metadata:
-           name: '{{ vm_name }}'
-           namespace: default
-           labels:
-             kubevirt.io/vm: test
-         spec:
-           running: false
-           template:
-             metadata:
-               labels:
-                 kubevirt-vm: cirros
-                 special: cirros-e2e-key
-             spec:
-               domain:
-                 resources:
-                   requests:
-                     memory: 512Mi
-                 cpu:
-                   cores: 1
-                 devices:
-                   disks:
-                   - volumeName: disk0-pvc
-                     name: mydisk
-                     disk:
-                       bus: virtio
-                   - name: cloudinitdisk
-                     volumeName: cloudinitvolume
-                     cdrom:
-                       bus: virtio
-               volumes:
-               - name: disk0-pvc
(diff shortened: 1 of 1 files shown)


tests/playbooks/e2e.yaml
```diff
@@ -41,52 +41,18 @@
...
-               - name: disk0-pvc
-                 persistentVolumeClaim:
-                   claimName: '{{ pvc_name }}'
-               - name: cloudinitvolume
-                 cloudInitNoCloud:
-                   userData: |
-                     #cloud-config
-                     password: "gocubsgo"
-                     chpasswd: ( expire: False )
+       memory: 64Mi
```
machacekondra: You don't need the cloud-init here?

vatsalparekh: the cirros image already has the cloudinit we're passing, so nothing specific about it
But I felt we should keep the `inline` object, so added it back now

vatsalparekh:
This might not work without #181 
Works fine with Python3

tests/playbooks/e2e.yaml
```diff
@@ -41,96 +41,56 @@
...
+        name: "{{ vm_name }}"
         namespace: default
 
     - name: Wait for VM to be ready
       k8s_facts:
         name: "{{ vm_name }}"
         namespace: default
         kind: VirtualMachine
-        api_version: kubevirt.io/v1alpha2
+        api_version: kubevirt.io/v1alpha3
```
pkliczewski: Why do we need `api_version` attribute here?

tests/playbooks/e2e.yaml
```diff
@@ -146,7 +106,7 @@
...
```
pkliczewski: same here. It looks like a bug to me

tests/playbooks/e2e.yaml
```diff
@@ -178,7 +138,7 @@
         name: "{{ vm_name }}"
         namespace: default
         kind: VirtualMachineInstance
-        api_version: kubevirt.io/v1alpha2
+        api_version: kubevirt.io/v1alpha3
```
pkliczewski: and here ...

tests/playbooks/e2e.yaml
vatsalparekh: Because `k8s_facts` needs it (or can it be avoided?)
https://github.com/ansible/ansible/blob/f49469f7bf42297ec47de16e7a5e851fbf1f442b/lib/ansible/modules/clustering/k8s/k8s_facts.py#L161

I tried to run without specifying the `api_version` and it just returns blank array, which is wrong

pkliczewski: OK, this means we need to simplify facts module as well. I will open a card to track it.

pkliczewski:
@machacekondra @mmazur review?

PR #181, opened 2019-02-12
Make api detection logic python2 compatible

mmazur:
Python3 vs. python2 incompatibilities made my last PR not work correctly on the latter. This should fix it.


pkliczewski:
@mmazur Please add a test

mmazur:
Ok, now it checks whether that version class actually works.

pkliczewski:
ATM, we are running [travis](https://github.com/kubevirt/ansible-kubevirt-modules/blob/master/.travis.yml#L4) only with py2. It would be great to add py3 as well.

PR #182, opened 2019-02-12
Limit reqs to openshift client <0.9.0

mmazur:
I've seen the changes in 0.9. Our code won't work with that version without some serious changes.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,4 +4,4 @@ mock >= 2.0.0, < 3.0.0
 flake8 >= 3.5.0, < 4.0.0
 ansible >= 2.7.0, < 3.0.0
 # Openshift client installable by pip:
-openshift >= 0.8.2, < 1.0.0
+openshift >= 0.8.2, < 0.9.0


pkliczewski:
Please provide more information what are the issues

mmazur:
Until it gets released and we commit to porting to it, there's no point in spending too much time investigating the current state of it. I did try using master openshift-restlclient-python when doing the api detection thing and one thing I can say is that it looks noticeably incompatible. For example `openshift.watch` (which we use in everything) doesn't exist anymore (`watch.py` file is like completely gone). And there were some definitions missing. Basically supporting it will be a separate case.

machacekondra:
Maybe better to open an issue on their repo that they break backward compatibility?

mmazur:
I'm pretty sure if something's versioned 0.X, it's because the author doesn't want to guarantee compatibility yet.

(This is btw explicitly stated on semver.org, item 4, which I've seen multiple references to over the past few months in various projects, so I'm assuming it has become a sort of standard for versioning stuff.)

pkliczewski:
We can open the issue and add comment to this change that it will be resolved in the future.

mmazur:
You mean:

1. add issue in our repo: add compatibility with openshift-restclient 0.9+
2. add link to this in commitlog

?

mmazur:
Done. Created an issue and added a link to the cilog.

pkliczewski:
@machacekondra review?

PR #186, opened 2019-02-13
Add datavolumes support

machacekondra:
Adding WIP so we can agree on design. Other than that it's working solution.

### DIFF
--- a/lib/ansible/module_utils/kubevirt.py
+++ b/lib/ansible/module_utils/kubevirt.py
@@ -200,6 +200,54 @@ class KubeVirtRawModule(KubernetesRawModule):
                 template_spec['volumes'] = []
             template_spec['volumes'].extend(spec_volumes)
 
+    def _define_datavolumes(self, datavolumes, spec):
+        """
+        Takes datavoulmes parameter of Ansible and create kubevirt API datavolumesTemplateSpec
+        structure from it
+        """
+        if not datavolumes:
+            return
+
+        spec['dataVolumeTemplates'] = []
+        for dv in datavolumes:
+            # Add datavolume to datavolumetemplates spec:
+            dvt = virtdict()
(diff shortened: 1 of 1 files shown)


machacekondra:
rebased

mmazur:
I think the proper way to do this is to integrate it the same way we've done pvc's. So this should work:

```
cpu_cores: 1
disks:
  - name: d1
    volume:
      (…)
    disk:
      bus: virtio
  - name: d2
    datavolume:
      all the datavolume-y definitions
    disk:
      bus: virtio
```

with all the magic required to get the datavolumes spec properly generated happening behind the scenes.

My brain fails me, so I'm unable to think through how the datavolume spec should potentially be flattened, but I suppose that's an option.

machacekondra:
I was afraid it would bring too much nesting. Like:
```yaml
kubevirt_vm:
  ...
  disks:
    - name: disk
      datavolume:
        source:
	  http:
	    url: https://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2
	pvc:
	  accessModes:
	    - ReadWriteOnce
	  storage: 2Gi
      disk:
        bus: virtio
```

But yes it's an option. Note that we do the same with cloud_init it's top level parameter as well, which create volume&disk. So I though we should put this top level as well because it creates those behind scene as well.

machacekondra:
Yeah, I will change this PR to test playbook and move code to Ansible repo.

pkliczewski:
> Yeah, I will change this PR to test playbook and move code to Ansible repo.

Do we need this PR if we want to push it to Ansible repo?

mmazur:
Yeah, we need to make sure to get everything pushed this week, even if it's not finished, to make sure the PR's are "old enough" to have a chance to get merged.

machacekondra:
Moved here: https://github.com/ansible/ansible/pull/52998

PR #192, opened 2019-02-14
Add e2e test to be executed per each new PR

machacekondra:
This PR add e2e test to be executed per each PR on travis on Minikube

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,35 @@
 sudo: required
 dist: xenial
 
+language: python
+python:
+  - "2.7"
+env:
+  - KUBECONFIG=tests/k8s/config
+
 branches:
   only:
   - master
   - "/^v[0-9]/"
-
+install:
+  - sudo pip install -U pyasn1 pyopenssl
+  - sudo pip install openshift --ignore-installed PyYAML
+  - sudo pip install -r requirements-dev.txt
 jobs:
   include:
     - stage: Tests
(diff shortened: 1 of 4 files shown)


machacekondra:
@pkliczewski @mmazur @vatsalparekh

.travis.yml
```diff
@@ -1,23 +1,38 @@
...
-  - KUBECONFIG=tests/k8s/config
-install:
-  - pip install -r requirements-dev.txt
-script:
-  - pytest --cov=lib tests
-  - flake8 --version
-  - flake8 lib/ansible/modules
+sudo: required
+dist: xenial
+
```
pkliczewski: Which version of python we will be using? I think we need to run py2 and py3.

mmazur: pytest from python3 will fail, so that'll warrant a separate PR to get all of that working

machacekondra: Ah, I accidentally removed uni tests, will re-add.

But more imporatant is to actually try to run both Ansible with Py3 and with Py2 I guess, I make a note and will add it. On f29 Py3 is default, not sure about Ubuntu, will check.

mmazur: I've checked it before and it complains, so adding py3 testing will require figuring out what's wrong. I suggest limiting this PR to what it says in this PR's title and then doing a separate 'add python3 compatibility and tests' push.

pkliczewski: Ack, will open an issue to enable py3

PR #198, opened 2019-02-15
Fix info on CA store

mmazur:
Previous info was wrong. This one is correct.


mmazur:
Afaik no. I did not look too deeply into it, but as far as I know openshift depends on kubernetes, which depends on certifi, so at least in our case this doc is true. Whether there's some other better mechanism for doing this stuff, I don't know.

mmazur:
Hmm, no, wait, this is more complicated.

mmazur:
Heh, seems `certifi` is the default choice, but for example Fedora seems to modify the `requests` package (as in – the python3-requests rpm) to make sure system ca bundle is used. This patch was supposed to be a one–liner, I'm not gonna be looking too deeply into this right now, I'll get to it later.

PR #201, opened 2019-02-20
Switch k8s_auth.yml to using module defaults

mmazur:
No more clunky yaml anchors, yay!

### DIFF
--- a/tests/playbooks/k8s_auth.yml
+++ b/tests/playbooks/k8s_auth.yml
@@ -3,6 +3,11 @@
 - name: Log onto OpenShift, do something, then log out
   hosts: localhost
   connection: local
+  module_defaults:
+    group/k8s:
+      host: https://192.168.42.55:8443
+      verify_ssl: no
+#      ssl_ca_cert: ca.pem
   vars:
     k8s_developer_password: zupa.8
   tasks:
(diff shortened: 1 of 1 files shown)


mmazur:
@pkliczewski @machacekondra @vatsalparekh plz mrg

tests/playbooks/k8s_auth.yml
```diff
@@ -13,34 +18,22 @@
...
-          ssl_ca_cert: "{{ k8s_auth_results.k8s_auth.ssl_ca_cert }}"
-          verify_ssl: "{{ k8s_auth_results.k8s_auth.verify_ssl }}"
-          api_key: "{{ k8s_auth_results.k8s_auth.api_key }}"
-
     # Previous task generated I(k8s_auth) fact, which you can then use
     # in k8s modules like this:
     - name: Get a list of all pods from any namespace
       k8s_facts:
-        <<: *k8s_auth_params
+        api_key: "{{ k8s_auth_results.k8s_auth.api_key }}"
```
machacekondra: Is it possible to have this as part of defaults as well?

mmazur: Only after I write a `set_module_default` module (analogous to `set_fact`), which I may or may not have already investigated. :)

For the time being: no. But it will be glorious once it's possible. :)

PR #205, opened 2019-02-26
Remove volumeName from tests

machacekondra:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1681888

### DIFF
--- a/tests/playbooks/kind_crud.yml
+++ b/tests/playbooks/kind_crud.yml
@@ -28,12 +28,10 @@
                       memory: 512M
                   devices:
                     disks:
-                    - volumeName: myvolume
-                      name: mydisk
+                    - name: myvolume
                       disk:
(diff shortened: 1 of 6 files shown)


tests/playbooks/kind_crud.yml
```diff
@@ -597,12 +587,11 @@
...
                   devices:
                     disks:
-                    - volumeName: myvolume
-                      name: mydisk
+                    - name: myvolume
                       disk:
                         bus: virtio
                     - name: cloudinitdisk
-                      volumeName: cloudinitvolume
+name: cloudinitvolume
```
pkliczewski: Actually this would break the yaml. Please align it correctly.

machacekondra: Thanks for noticing. Done.

PR #214, opened 2019-03-22
Use minishift in CI

machacekondra:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "ci"]
 	path = ci
-	url = https://github.com/fabiand/traviskube
+	url = https://github.com/mmazur/traviskube
+	branch = providers
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,26 +12,30 @@ install:
   - sudo pip install -r requirements-dev.txt
 jobs:
   include:
-    - stage: Tests
-      name: "Default (minikube with latest k8s)"
+    - name: "minishift"
+      env:
+        - CPLATFORM=minishift
+        - KUBEVIRT=v0.17.0
+        - USE_OPERATOR=yes
+        - ENABLE_DATAVOLUMES=yes
 
 cache:
   - pip
   - directories:
     - cache
-    - "~/.minikube/cache"
+    - "~/.minishift/cache"
 before_script:
 ## FIXME Workaround for https://github.com/kubernetes/kubernetes/issues/61058
 ### And https://github.com/LiliC/travis-minikube/blob/e0f26f7b388057f51a0e2558afd5f990e07b6c49/.travis.yml#L11
 - sudo mount --make-rshared /
 
 - bash -x ci/prepare-host $CPLATFORM
-- bash -x ci/prepare-host virtctl
-- bash -x ci/start-cluster $CPLATFORM $CVER
-- bash -x ci/deploy-kubevirt $CPLATFORM
+- bash -x ci/prepare-host virtctl $KUBEVIRT
+- bash -x ci/start-cluster $CPLATFORM
+- bash -x ci/deploy-kubevirt $CPLATFORM $KUBEVIRT
 
 script:
-- timeout 20m bash -x test.sh
+- timeout 30m bash -x test.sh
 
 notifications:
   email:
--- a/test.sh
+++ b/test.sh
@@ -1,18 +1,14 @@
 #!/bin/bash
 
-echo "$(minikube ip) minikube" | sudo tee -a /etc/hosts
-
-## No HW virt
-kubectl create configmap -n kubevirt kubevirt-config --from-literal debug.useEmulation=true
-kubectl scale --replicas=0 deployment/virt-controller -n kubevirt
-kubectl scale --replicas=2 deployment/virt-controller -n kubevirt
+echo "$(minishift ip) minishift" | sudo tee -a /etc/hosts
 
 ## Prepare CDI
-export CDI_VERSION=v1.5.0
-kubectl apply -f https://github.com/kubevirt/containerized-data-importer/releases/download/$CDI_VERSION/cdi-controller.yaml
+export CDI_VERSION=$(curl https://github.com/kubevirt/containerized-data-importer/releases/latest | grep -o "v[0-9]\.[0-9]*\.[0-9]*")
+oc create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$CDI_VERSION/cdi-operator.yaml
+oc create -f https://github.com/kubevirt/containerized-data-importer/releases/download/$CDI_VERSION/cdi-operator-cr.yaml
 
 get_remaining_pods() {
-  kubectl get pods \
+  oc get pods \
     --all-namespaces \
     --field-selector=status.phase!=Running,status.phase!=Succeeded ;
 }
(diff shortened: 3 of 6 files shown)


tests/playbooks/e2e.yaml
```diff
@@ -92,12 +92,6 @@
             protocol: TCP
         selector:
           special: cirros-e2e-key
-        definition:
-          metadata:
-            labels:
-              kubevirt.io/vm: cirros-e2e-service
-          spec:
-            externalTrafficPolicy: Cluster
```
vatsalparekh: Not sure why to remove this?

]
```diff
@@ -0,0 +1,27 @@
...
+# .       message (or the oneline, if no original merge commit was
+# .       specified). Use -c <commit> to reword the commit message.
+#
+# These lines can be re-ordered; they are executed from top to bottom.
+#
+# If you remove a line here THAT COMMIT WILL BE LOST.
+#
+# However, if you remove everything, the rebase will be aborted.
+#
+# Note that empty commits are commented out
```
Comment (vatsalparekh) [inline ]]: @machacekondra added by mistake?

mmazur:
I've bumped the `kubevirt_vm` call that times out to have a timeout value of 5 minutes instead of 2 minutes and it still fails. Looks like something else is wrong. Let's see…

mmazur:
I'm pretty sure I managed to get the vms running by messing a bit with how kubevirt is getting deployed, but now I've hit an interesting problem. When an ephemeral vm exists and is running (I'm guessing this isn't just limited to ephemeral vms), the task below should return success with changed==false.

```
kubevirt_vm:
  state: present
  name: xyz
  ephemeral: yes
```

And it does on kubevirt 1.6.0 running on vanilla k8s 1.13.3. But it doesn't on same kubevirt running on latest minishift (or not so latest minishift). And this is the exact issue currently reported by travis. I'm suspecting some weird behavior on the openshift lib layer.

mmazur:
Blocked by: https://github.com/ansible/ansible/issues/55817

mmazur:
Ugh, so openshift 3.11 ships kubernetes 1.11 and there's a bug in that k8s that only got fixed in 1.12 (https://github.com/kubernetes/kubernetes/pull/67562), which is what we're hitting here.

I'll see if I can get the ansible/k8s people to add a workaround for this (ideally before 2.8 ships).

mmazur:
:angry: !#@$!^#$ :rage: 

Works… kind of. I'm pretty sure this will be even more flaky than the minkube environment, so expect random stuff to time out for no reason. Might be a problem for us, might not. If it is, we can investigate something other than travis at some later time.

For now, @machacekondra @pkliczewski @vatsalparekh please review and merge.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "ci"]
 	path = ci
-	url = https://github.com/fabiand/traviskube
+	url = https://github.com/mmazur/traviskube
```
pkliczewski: I understand there are bugs. Please paste reference or describe why we are using our fork.

.travis.yml
```diff
@@ -12,26 +12,30 @@ install:
   - sudo pip install -r requirements-dev.txt
 jobs:
   include:
-    - stage: Tests
-      name: "Default (minikube with latest k8s)"
+    - name: "minishift"
+      env:
+        - CPLATFORM=minishift
+        - KUBEVIRT=v0.16.0
```
pkliczewski: v0.16.2 - latest released

tests/playbooks/kubevirt_vm.yml
```diff
@@ -33,13 +38,15 @@
...
+        wait: no
       with_items:
         - { name: 'eph1', state: 'present' }
         - { name: 'eph1', state: 'running' }
         - { name: 'eph2', state: 'present' }
         - { name: 'eph2', state: 'running' }
       register: result
-      failed_when: result.changed != false or result.kubevirt_vm.status.phase != 'Running'
+      failed_when: result.kubevirt_vm.status.phase != 'Running'
+#     failed_when: result.changed != false or result.kubevirt_vm.status.phase != 'Running'
```
pkliczewski: why do we need this commented line?

tests/playbooks/kubevirt_vm.yml
```diff
@@ -58,13 +65,15 @@
...
+        wait: no
       with_items:
         - { name: 'eph1', state: 'stopped' }
         - { name: 'eph1', state: 'absent' }
         - { name: 'eph2', state: 'stopped' }
         - { name: 'eph2', state: 'absent' }
       register: result
-      failed_when: result.changed != false or 'spec' in result.kubevirt_vm or 'status' in result.kubevirt_vm
+      failed_when: ('spec' in result.kubevirt_vm or 'status' in result.kubevirt_vm)
+#     failed_when: result.changed != false or 'spec' in result.kubevirt_vm or 'status' in result.kubevirt_vm
```
pkliczewski: Same here

tests/playbooks/kubevirt_vm.yml
```diff
@@ -92,13 +101,15 @@
...
+        wait: no
       with_items:
         - { name: 'vm1', state: 'stopped' }
         - { name: 'vm1', state: 'present' }
         - { name: 'vm2', state: 'stopped' }
         - { name: 'vm2', state: 'present' }
       register: result
-      failed_when: result.changed != false or 'status' in result.kubevirt_vm
+      failed_when: ('status' in result.kubevirt_vm)
+#     failed_when: result.changed != false or 'status' in result.kubevirt_vm
```
pkliczewski: I see there are more. We use version control not to keep things like this.

machacekondra:
I would rather prefer stable tests without testing templates, then unstable env with templates testing.

mmazur:
It's my guess that the reliability issues we've had with the test env were due to k8s/kubevirt and non-accelarated vms in particular being rather heavy, so whenever there was a cpu shortage we got random timeouts. 

And since k8s is lighter than openshift I suspect that might be more of a problem. 

But I don't know for sure and this patch is already done. If it turns out to be a problem, we can revert this easily or consider moving away from travis. Or maybe it'll be enough to increase a few timeout values. I'll keep an eye on it going forward.

.gitmodules
mmazur: My changes are very trivial, and update here and there. The only reason is the RH Summit. Fabian won't be looking at low prio PRs for some time and I didn't want to get stuck on this. As soon Fabian is back to his usual responsive self I'm planning on getting my changes merged and reverting this back to his repo.

.travis.yml
mmazur: 0.16.1 and 0.16.2 are tagged as pre–release, which is why I didn't use them. Dunno if those tags can be ignored or if they actually mean something.

tests/playbooks/kubevirt_vm.yml
mmazur: There's a 4 line explanation of what this is for at the start of the playbook.

I did it this way on purpose. If I'm making a temporary workaround which I want to get removed as soon as possible, then I want to very clearly mark all the places in the code where the workaround applies. The only chance for someone to update it in a few months so that full tests are utilized is to make it an eye sore and hard to not notice should anyone look at this file.

You know how this stuff works. If I hide these lines in git, then no one will ever bother to try to fish out the originals and figure out what to do with them, so the "workaround" will just become the permanent tests until the heat death of the universe. But if the originals are still there in plain sight, there's a chance…

.gitmodules
vatsalparekh: Maybe we can move it under kubevirt org

.travis.yml
pkliczewski: There is `v0.17.0` available please change.

test.sh
```diff
@@ -1,18 +1,14 @@
 #!/bin/bash
 
-echo "$(minikube ip) minikube" | sudo tee -a /etc/hosts
-
-## No HW virt
-kubectl create configmap -n kubevirt kubevirt-config --from-literal debug.useEmulation=true
```
machacekondra: We don't need this anymore?

machacekondra: Ah, ok, you moved most of this to your repo.

mmazur:
I'll open a separate PR that adds jenkins support.

PR #215, opened 2019-03-22
Switch to 'wait: yes' for PVCs, part 2

mmazur:
I missed a spot.

### DIFF
--- a/tests/playbooks/e2e.yaml
+++ b/tests/playbooks/e2e.yaml
@@ -20,16 +20,7 @@
         cdi_source:
           http:
             url: "https://download.cirros-cloud.net/0.4.0/cirros-0.4.0-x86_64-disk.img"
-
-    - name: Wait for PVC to be ready
-      k8s_facts:
-        name: "{{ pvc_name }}"
-        kind: PersistentVolumeClaim
-        api_version: v1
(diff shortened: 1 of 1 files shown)


tests/playbooks/e2e.yaml
```diff
@@ -20,16 +20,7 @@
...
-    - name: Wait for PVC to be ready
-      k8s_facts:
-        name: "{{ pvc_name }}"
-        kind: PersistentVolumeClaim
-        api_version: v1
-      register: pvc
-      until: "pvc.resources[0].status.phase == 'Bound' and 'cdi.kubevirt.io/storage.pod.phase' in pvc.resources[0].metadata.annotations and pvc.resources[0].metadata.annotations['cdi.kubevirt.io/storage.pod.phase'] == 'Succeeded'"
-      delay: 15
-      retries: 30
+        wait: yes
```
pkliczewski: Do we want to customize timeout?

mmazur: The default is quite high (5 minutes), as this is PVCs, so it's expected to take a while (especially with CDI involved). Tests passed, so I assume we're fine for now. If it takes more than 5 minutes, we likely have an issue with the CI.

PR #216, opened 2019-03-27
[WIP] Rely on wait logic in kubevirt_vm

mmazur:
Once https://github.com/ansible/ansible/pull/54404 is merged, we should be able to rely on kubevirt_vm's built–in wait logic. And test it more.

### DIFF
--- a/tests/playbooks/e2e.yaml
+++ b/tests/playbooks/e2e.yaml
@@ -7,7 +7,9 @@
       namespace: default
   vars:
     vm_name: cirros-e2e-vm
+    vmi_name: cirros-e2e-vmi
     pvc_name: cirros-e2e-pvc
+    pvc2_name: cirros-e2e-pvc2
     service_name: cirros-e2e-service
   tasks:
     - name: Create PVC with CDI annotations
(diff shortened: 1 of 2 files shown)


tests/playbooks/e2e.yaml
```diff
@@ -22,6 +24,50 @@
...
+            bus: virtio
+
+    - name: Verify the ephemeral VM is in fact running
+      k8s_facts:
+        name: "{{ vmi_name }}"
+        kind: VirtualMachineInstance
+        api_version: kubevirt.io/v1alpha3
+      register: vmi
+      until: "vmi.resources[0].status['phase'] == 'Running'"
+      delay: 1
```
pkliczewski: Why do we need this so low? Do we assume that vm is always there if so why do we need it here?

mmazur: Because this isn't supposed to be looped wait logic anymore, as the wait logic is inside  kubevirt_vm now. This is supposed to be more of an assert verifying that the wait logic inside kubevirt_vm worked correctly and the VMI is really ready (ditto for the rest of similar instances).

This is just WIP, so I wanted to "annotate" all the relevant parts of the playbooks for myself and this was the quickest way to do it (near instant loop with little to no repeats == almost an assert). I'll likely replace this non–loopy loop with a proper k8s_facts and an assert at some later time.

PR #218, opened 2019-04-03
Add comprehensive kubevirt_vm testing

mmazur:
Make sure all the logic gets tested.


mmazur:
And done.

tests/playbooks/all.yml
```diff
@@ -1,4 +1,5 @@
-- import_playbook: e2e.yaml
+- import_playbook: kubevirt_vm.yml
```
pkliczewski: Shouldn't we keep this in alphabetical order?

mmazur: I though we'd start with the more specific per–module tests first and if those pass, leave `e2e.yaml` for a more complex mix of everything.

PR #227, opened 2019-05-21
Switch to oVirt CI for (most of) the tests

mmazur:
If oVirt CI is slow enough, it makes sense to keep travis around to do some basic install/syntax checking.


mmazur:
This is of course WIP for now, but I'm not sure if labeling it as such wouldn't disable CI?

pkliczewski:
You can add `[WIP]` to the PR title and it should be clear. How do you want to narrow list of tests run in travis?

mmazur:
I've limited it to the only thing that I could for now, which is running `ansible-playbook --check-syntax`, as that will happen much quicker than getting to the same step on the jenkins build. If we ever have any other ideas for tests, we can always extend it.

ovirt-infra:
Hello contributor, thanks for submitting a PR for this project! 

I am the bot who triggers "standard-CI" builds for this project.
As a security measure, I will not run automated tests on PRs that are not from white-listed contributors.

In order to allow automated tests to run, please ask one of the project maintainers to review the code and then do one of the following:

1. Type `ci test please` on this PR to trigger automated tests for it.
2. Type `ci add to whitelist` on this PR to trigger automated tests for it and also add you to the contributor white-list so that your future PRs will be tested automatically. ( keep in mind this list might be overwritten if the job XML is refreshed, for permanent whitelisting, please follow #3 option )
3. If you are planning to contribute to more than one project, maybe it's better to ask them to add you to the project organization, so you'll be able to run tests for all the organization's projects.

mmazur:
@pkliczewski please say `ci add to whitelist`

pkliczewski:
ci add to whitelist

mmazur:
Works. This is an almost 1:1 replacement of our travis+minikube solution. Two things are still left unsolved by this patch:

1. I had to disable one test in e2e.yml. KubevirtCI doesn't (yet) have a mechanism for exposing any of the cluster's networking to the host OS, so trying to tcp connect to a port on a VMI isn't doable. I have talked to Roman about it a while ago and he said they're open to patches. I did look into the issue then and it's non–trivial, so I'd say we just leave it for now.
2. The goal was getting the Template tests working and this doesn't do that. I'd like to do that in a separate patch, ideally going OKD4, so we don't have to switch to that a few months from now. However support for that is an ongoing issue, so it might make sense to give it a few more weeks. Again, I'll start with bugging Roman about it.

mmazur:
Yeah, just did a quick test, the okd4 support in kubevirtci isn't usable yet, at least for what we need. So the question is do we stick with vanilla k8s and revisit in a few weeks or just go os 3.11 for the time being?

pkliczewski:
I would keep working with k8s and once ready moved to okd4. There is no need to waste more time on it.

mmazur:
Ok, I know what else we can use for the `e2e.yaml` thing is – readiness probes on our vms and we'll get the same effect. I'll do that in a separate patch.

WRT okd4 or openshift in general, I'll take a look later if I can hammer out a quick switch to openshift3, which would get us what we need. If not, we should probably wait for okd4, which, judging from the ongoing discussions, is not something only we'd like to see, so it will happen eventually.

So, please review and merge.

tests/playbooks/e2e.yaml
```diff
@@ -99,22 +99,22 @@
           spec:
             externalTrafficPolicy: Cluster
 
-    - name: Gather facts for {{ vm_name }} VMI in default namespace
-      k8s_facts:
-        name: "{{ vm_name }}"
-        kind: VirtualMachineInstance
-        api_version: kubevirt.io/v1alpha3
-      register: vmi
+#   - name: Gather facts for {{ vm_name }} VMI in default namespace
```
pkliczewski: I do not like that we are commenting the code. What are the reasons that it is not working?

mmazur: Architectural differences between minikube and kubevirtci. Today I've been told of a way to reenable the test in an alternative form, which I was planning on doing in a separate PR.

mmazur:
Occurred to me that this could probably use a short intro:

1. `automation/test.sh` is the actual test, very short file
2. `Makefile` and `cluster/*` set up (and tear down) the actual cluster; it's equivalent to the `ci/` dir of old
3. Everything else is fluff/temporary workarounds

PR #236, opened 2019-05-28
Add openshift3 jenkins target support

mmazur:
Fingers crossed…

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ kubernetes
 *.swp
 ansible.log
 .kubeconfig*
+cluster/oc
+cluster/kubectl
--- a/automation/check-patch.os-3.11.0.mounts
+++ b/automation/check-patch.os-3.11.0.mounts
@@ -1,0 +1,1 @@
+check-patch.mounts
--- a/automation/check-patch.os-3.11.0.packages
+++ b/automation/check-patch.os-3.11.0.packages
@@ -1,0 +1,1 @@
+check-patch.packages.el7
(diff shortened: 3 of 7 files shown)


mmazur:
Right, so here's the deal:
1. This patch adds openshift3* target support.
2. As can be seen from the CI output it does **not** break current k8s vanilla target.
3. But the openshift target doesn't actually work for some bizarre reason and I don't have the time to debug it right now.

So, what I'd like to do is just drop the openshift target from the travis config (so only k8s vanilla gets built), but keep all the other changes. That way someone maybe might try to debug this some time from now or use it as basis for okd4 support.

@machacekondra @pkliczewski if you're fine with that, I'll push this again with the modified travis config and then it can be merged. K?

pkliczewski:
@mmazur sounds like good plan

mmazur:
@vatsalparekh sorry, but 0templates4you. According to the [build log](https://kubevirt-master-stdci-production.apps.ovirt.org/blue/organizations/jenkins/kubevirt_ansible-kubevirt-modules_standard-check-pr/detail/kubevirt_ansible-kubevirt-modules_standard-check-pr/49/pipeline/92):

1. Cluster went up fine.
2. Kubevirt and CDI deployed no problem (line 377).
3. Ansible can't connect to the cluster due to… ssl handshake problems? (line 1236) Beats me what that's about. `oc` has no issues connecting.

Anyway, this won't happen any time soon.

mmazur:
Ok, this works, please merge.

PR #239, opened 2019-05-30
Switch to using 4 nodes on Jenkins

mmazur:
Let's see if it's faster with the tests this way (and less prone to timeouts). The machines can handle this easily, they have 100+ gigs of ram and like 16 threads.


mmazur:
ci test please

mmazur:
ci test please

Issue #242, opened 2019-05-31
ansible @devel and @stable-2.8 break kubevirt_pvc (annotations aren't set)


mmazur:
You know how CI has been flaky recently?

One reason is that #240 is actually a real thing that we're hitting on occasion. (But that's a different story.)

The other is that CI is using devel branch of ansible (on purpose, to catch stuff early) and that actually broke in recent days. And that's the other bug we've been hitting.

Specifics are: if trying to utilize CDI with `kubevirt_pvc`, that depends on setting annotations on the PVC. But with most recent devel ansible, annotations aren't set, so the whole thing times out.

This warrants an investigation on Monday. I'll also push a PR against our tests to increase verbosity, so that this becomes easier to notice next time.

/kind bug

pkliczewski:
Tracked by BZ [1716905](https://bugzilla.redhat.com/1716905)

mmazur:
Github logic was mistaken.

mmazur:
Fix merged both in stable-2.8 and devel branches. 2.8.2 should be released in a few days. Closing.

PR #243, opened 2019-05-31
Better logging + test multiple ansible versions

mmazur:
This will make it waay easier to figure out what's causing CI failures.


mmazur:
Ok, this PR works exactly as expected. It also confirms that ansible stable works perfectly fine with our current tests, while ansible devel fails on pvc/cdi (aka issue #242).

Please merge @machacekondra @pkliczewski

PR #244, opened 2019-06-04
Add vm_image role

### LINKED ISSUE #242: ansible @devel and @stable-2.8 break kubevirt_pvc (annotations aren't set)

mmazur:
No tests, because:

1. Role tests aren't enabled.
2. And even if they were, #242 wouldn't let this role pass correctly anyway.

Maybe we can deal with enabling role tests as part of the "fix #242" story, next sprint? Or do I just plunk some test somewhere and we'll fix it later?

### DIFF
--- a/roles/vm_image/examples/example2.yml
+++ b/roles/vm_image/examples/example2.yml
@@ -1,0 +1,25 @@
+---
+- name: Create volumes from local files
+  hosts: localhost
+  module_defaults:
+    # Hints to the upload module on where to connect to
+    kubevirt_cdi_upload:
+      upload_host_verify_ssl: false
+      upload_host: https://127.0.0.1:8443
+
+  vars:
+    kubevirt_image_volumes:
+      - name: volume1
+        namespace: default
+        size: 100Mi
+        source: "/tmp/cirros-0.4.0-x86_64-disk.img"
+
+      - name: volume2
+        namespace: default
+        size: 100Mi
+        source: "file:///tmp/cirros-0.4.0-x86_64-disk.img"
+
+
+  roles:
+    - kubevirt.vm_image
+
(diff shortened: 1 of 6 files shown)


roles/vm_image/examples/example2.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Create volumes 2
+  hosts: localhost
+  module_defaults:
```
pkliczewski: The examples are really similar. What do we want to show in both? Do you think it would be good idea to have only one example?

machacekondra:
> No tests, because:
>
> Role tests aren't enabled.

They were, why they are not now?

roles/vm_image/examples/example2.yml
mmazur: Hmm, the only reasoning for the second file was to showcase the `kubevirt_cdi_upload` upload host override. I'll drop the other examples from that file and just leave the upload one.

mmazur: Done.

pkliczewski:
@mmazur we still need to work on `CreatePVCFailed: PVC creation timed out`

PR #247, opened 2019-06-05
Disable tests of known broken modules

mmazur:
Until we have separate jobs for devel and stable ansible.

### DIFF
--- a/tests/playbooks/all.yml
+++ b/tests/playbooks/all.yml
@@ -1,6 +1,8 @@
 - import_playbook: kubevirt_pvc.yml
+  when: ansible_version.full < "2.9.0"
 - import_playbook: kubevirt_vm.yml
 - import_playbook: kubevirt_dv_vm.yaml
 - import_playbook: kubevirt_preset.yml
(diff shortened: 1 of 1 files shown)


automation/test.sh
```diff
@@ -27,6 +27,9 @@ pip3 install --user ansible
 ansible-playbook --version
 ansible-playbook -vvv tests/playbooks/all.yml
 
+# TODO: REMOVE THIS LATER
```
pkliczewski: This seems to prevent us to get any error. It would be better not to test broken code for devel branch only.

machacekondra: Yeah, can we just use github API to fetch issue and skip the test if the issue is open and don't skip if it's closed, that way we now it was fixed properly and we will be testing everything working always

mmazur:
Ok, this patch disables the borked tests on dev ansible. I need to do other stuff first, before I can add the issue–detection logic, so please merge this for now and I'll send the rest in another PR.

pkliczewski:
@machacekondra what do you think?

PR #249, opened 2019-06-07
Dynamic test skipping based on issues status

mmazur:
The case detection logic can't be done globally in an easy manner, since there's no variable sharing between playbooks. So it's run for each playbook with known potential issues. Seems to work.

### DIFF
--- a/tests/playbooks/known_issues.yaml
+++ b/tests/playbooks/known_issues.yaml
@@ -1,0 +1,28 @@
+- name: Set some global known_issues vars
+  set_fact:
+    skip_pvc: false
+    skip_vm: false
+    skip_dv_vm: false
+    skip_preset: false
+    skip_vmir: false
+    skip_e2e: false
+    current_playbook: ''
+
+    github_issues:
+    - name: pvc_module_242
+      url: https://api.github.com/repos/kubevirt/ansible-kubevirt-modules/issues/242
+
+- name: Fetch github issues
+  include_tasks: known_issues/github.yaml
+  with_items: "{{ github_issues }}"
+
+- name: "Known issue: ansible.git@devel breaks kubevirt_pvc #242"
+  set_fact:
+    skip_pvc: true
+    skip_e2e: true
+  when: (ansible_version.full == "2.8.1" or ansible_version.full == "2.8.2") and pvc_module_242.state == "open"
+
+- name: "Skipping this playbook due to open issues"
+  meta: end_play
+  when: "skip_{{ current_playbook }} == true"
+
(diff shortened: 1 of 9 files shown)


mmazur:
ci test please

tests/playbooks/known_issues.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+- name: Fetch github issues
+  include_tasks: known_issues/github.yaml
+  with_items: "{{ github_issues }}"
+
+- name: "Known issue: ansible.git@devel breaks kubevirt_pvc #242"
+  set_fact:
+    skip_pvc: true
+    skip_e2e: true
+  when: (ansible_version.full == "2.9.0" or ansible_version.full == "2.8.1") and pvc_module_242.state == "open"
```
pkliczewski: Why to depend on specific z-stream? Do we want to update this number whenever new release is available?

tests/playbooks/kubevirt_pvc.yml
```diff
@@ -3,6 +3,11 @@
   hosts: localhost
   connection: local
   tasks:
+    - name: Check for known open issues
```
pkliczewski: Should we add this to every tests playbook? I see that we defined more variables in `known_issues.yaml`

mmazur: I wanted to make it run only once globally, but ansible doesn't allow that. So it needs to be included at the beginning of each playbook. It made sense to me to only include it in playbooks that are known to have open issues, but I guess it doesn't really matter if it runs for every playbook, it's just one http call. I'll put it everywhere.

pkliczewski: I would added it everywhere so we can easily track issues related to other modules.

tests/playbooks/known_issues.yaml
mmazur: When a new release is available, we'll remove the whole test. We (will) test against:
- released (pip install ansible) – 2.8.1
- stable branch – 2.8.2 (to be released)
- devel branch – 2.9.0 (to be released)

If any of those have a known issue, then we need to skip certain tests, but *only* for affected branch/release. At present, the affected versions for this case are: released 2.8.1 and stable-2.8 branch. And I'm gonna put only those two versions for skipping, since I do want to run all of my tests on devel.

mmazur:
Ha, it works. Please merge.

mmazur:
Nope, the ==2.8.2 is wrong, that's not what `stable-2.8` ansible introduces itself as. I'll  look into it later.

PR #250, opened 2019-06-13
Move non-test playbooks into examples/ dir

mmazur:
It's confusing with too many yamls, when only a few of those are actual tests.


mmazur:
ci test please

mmazur:
ci test please

PR #251, opened 2019-06-13
Bring back travis builds based on minikube

mmazur:
Travis is faster/more reliable than our Jenkins, if more limited in what it can do.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "traviskube"]
+	path = travis/traviskube
+	url = https://github.com/fabiand/traviskube
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,33 +1,49 @@
-sudo: required
+notifications:
+  email:
+    recipients:
+      - mmazur@redhat.com
+      - omachace@redhat.com
+    on_success: never
+    on_failure: always
+
 dist: xenial
+sudo: required
+language: python
+python:
+  - "3.7"
 
 branches:
   only:
-  - master
-  - "/^v[0-9]/"
-install:
-  - sudo pip install -U pyasn1 pyopenssl
-  - sudo pip install openshift --ignore-installed PyYAML
-  - if [ -n "$PR_NUMBER" ]; then sudo pip install git+https://github.com/ansible/ansible.git@refs/pull/$PR_NUMBER/head; else sudo pip install git+https://github.com/ansible/ansible.git; fi
-  - sudo pip install -r requirements-dev.txt
-jobs:
-  include:
-    - stage: Tests
-      name: "Default (Playbook syntax test)"
+    - master
 
 cache:
   - pip
   - directories:
     - cache
-before_script:
-script:
-- ansible-playbook --version
-- ansible-playbook -vvv --syntax-check tests/playbooks/all.yml
+    - "~/.minikube/cache"
 
-notifications:
-  email:
-    recipients:
-      - omachace@redhat.com
-      - mmazur@redhat.com
-    on_success: never
-    on_failure: always
+# If you trigger a custom build with a custom config like this:
+#   env: ANSIBLE_SOURCE=pr/9999
+# then ansible from ansible.git's PR 9999 will be installed and used for executing the playbook
+env:
+  - ANSIBLE_SOURCE=pip
+  - ANSIBLE_SOURCE=branch/stable-2.8
+  - ANSIBLE_SOURCE=branch/devel
+
+install:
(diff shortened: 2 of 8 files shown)


mmazur:
@pkliczewski @machacekondra ready for review and merge (hopefully all the CI is green by the time it finishes :).

Note on dropping the dynamic github issue checking. Two reasons:
1. We're testing three different branches. We'd potentially need three different issues for each bug to actually control tests with issues. Otherwise someone needs to modify the `known_issues.yaml` anyway, and if we have to do that, then the issue checking is superfluous.
2. More importantly, github has api rate limiting per ip, which makes the api unusuable from travis hosts. And the only way around it would be to add someone's github token to `.travis.yaml`, which obviously won't happen. (Adding the token as a project variable in travis isn't an option, since those don't get exported to externally–pushed PRs, which is what we use exclusively.)

pkliczewski:
@mmazur We could solve the issues by:

> 1. We're testing three different branches. We'd potentially need three different issues for each bug to actually control tests with issues. Otherwise someone needs to modify the `known_issues.yaml` anyway, and if we have to do that, then the issue checking is superfluous.

Using labels which would specify for which branch the issue was open. In this way we will have control over which issues should skip which tests.

> 2. More importantly, github has api rate limiting per ip, which makes the api unusuable from travis hosts. And the only way around it would be to add someone's github token to `.travis.yaml`, which obviously won't happen. (Adding the token as a project variable in travis isn't an option, since those don't get exported to externally–pushed PRs, which is what we use exclusively.)

We could use kubevirt bot credentails to interact with the api.

mmazur:
> @mmazur We could solve the issues by:
> 
> > 1. We're testing three different branches. We'd potentially need three different issues for each bug to actually control tests with issues. Otherwise someone needs to modify the `known_issues.yaml` anyway, and if we have to do that, then the issue checking is superfluous.
> 
> Using labels which would specify for which branch the issue was open. In this way we will have control over which issues should skip which tests.

Yup, that would work, especially considering the standardized `ANSIBLE_SOURCE` target names in this PR. I just created a label called `affects=branch/stable-2.8`, so the test would be something like `if state == open and "affects=$ANSIBLE_SOURCE" in labels` (assuming ansible/jinja2 have enough dict/list transformation functions in place to be able to test for this).

> > 1. More importantly, github has api rate limiting per ip, which makes the api unusuable from travis hosts. And the only way around it would be to add someone's github token to `.travis.yaml`, which obviously won't happen. (Adding the token as a project variable in travis isn't an option, since those don't get exported to externally–pushed PRs, which is what we use exclusively.)
> 
> We could use kubevirt bot credentails to interact with the api.

No, the problem isn't with whose credentials we use. Issue is that for security reasons travis does not allow using the token in PR builds (because a PR can contain anything, including code to just mail the token to an external attacker). In other words:
1. Scheduled/manual/triggered builds of any tag/branch from kubevirt/ansible-kubevirt-modules.git? You get your GITHUB_TOKEN no problem.
2. Somebody pushes a PR from their own fork of the repo (like we do all of the time to get any changes in)? No token for you, hence no github issue detection.

Only way around this I can think of (except just putting the token in `travis.yml`) is to cache the status of the issue as obtained in type 1 builds and have type 2 builds use that cached issue instead of accessing github's api. It might work.

Though between the caching support (and having it work correctly on both jenkins and travis) and adding the labels, this becomes quite a task of its own. I'm pretty sure it's not worth the effort. The test skipping logic might come in handy once in a while, but (hopefully) very rarely and we should be fine just using PRs to control it, without needing the issue fetching.

pkliczewski:
> 2. Somebody pushes a PR from their own fork of the repo (like we do all of the time to get any changes in)? No token for you, hence no github issue detection.
> 
> Only way around this I can think of (except just putting the token in `travis.yml`) is to cache the status of the issue as obtained in type 1 builds and have type 2 builds use that cached issue instead of accessing github's api. It might work.
> 
> Though between the caching support (and having it work correctly on both jenkins and travis) and adding the labels, this becomes quite a task of its own. I'm pretty sure it's not worth the effort. The test skipping logic might come in handy once in a while, but (hopefully) very rarely and we should be fine just using PRs to control it, without needing the issue fetching.

This sounds too complex and not worth spending time on. I think we need to step back and check how to fix our current issues without spending too much time on it.

mmazur:
The playbook–skipping without issue checking is good enough for the rare cases that we'll have regressions. IOW please look at how pretty and clear the Travis CI output is with this and merge it.

(Travis is set to rerun every 24h, so once merged, this will give us very quick warnings on any borked branches.)

PR #254, opened 2019-06-18
Add topology_key to affinity tests and examples

machacekondra:
This PR fixes the affinity tests of related to app_deployment role. `topology_key` is required, so it must be passed in tests as well.


mmazur:
@machacekondra please rebase on master, so that the role tests are actually run by travis.

machacekondra:
Fixed

Issue #240, opened 2019-05-31
Repeated CI instability on minikube (travis)


mmazur:
/kind bug

We've been hitting a random issue when running our ~~pvc creation~~ tests, which ends up with this error message:

```
Failed to create object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"Internal error occurred: resource quota evaluates timeout\",\"reason\":\"InternalError\",\"details\":{\"causes\":[{\"message\":\"resource quota evaluates timeout\"}]},\"code\":500}\\n'"
```

Since it wasn't consistent, we've chalked it up to TravisCI being weird. Now we're on Jenkins with completely different hw running the cluster+tests and are seeing the same issue.

Google says that this is rare, but does happen on occasion:

1. https://github.com/gluster/gluster-csi-driver/issues/92
2. https://github.com/kubernetes/kubernetes/issues/67531 

Two scenarios come to my mind at this time:

1. K8S has performance knobs that can be tuned and the https://github.com/kubevirt/kubevirtci cluster we're using has them turned down too much and we're on occasion getting some parts of the cluster (etcd?) to choke on the workload.
    - [ ] Possible action: ask kubevirtci people about their take on this
2. This is simply an obscure bug and we're unlucky.
    - [x] ~~Action1: mess around with task ordering in our test playbooks to maybe help avoid hitting the issue.~~ Tried, didn't help, it just happens on a different task now. (See comments.)
    - [ ] Action2: get our jenkins mock-runner to endlessly loop the tests on my own hw and hope to hit the bug, at which point I can get debug info out of it.
    - [ ] Action3: connect with our CI people and get them to somehow freeze a CI machine for further debugging next time we hit the issue.

Suggestions welcome.

mmazur:
Btw, the timeout on this is only 10 seconds: https://github.com/kubernetes/kubernetes/blob/cc67ccfd7f4f0bc96d7f1c8e5fe8577821757d03/plugin/pkg/admission/resourcequota/controller.go#L620

pkliczewski:
@mmazur can you check how kubevirt ci is working? I think they should have solved the issues already. Let's reuse what they did so far.

mmazur:
@davidvossel has started working on a consumable version of the cluster setup/teardown scripts just now. As soon as they're ready, we'll switch. If it fixes this issue, great. If not, then it won't be just our problem. :)

[...]

PR #255, opened 2019-06-18
Switch to upstream kubevirtci scripts for cluster setup

mmazur:
1. Switching to bats was done because bats' output might be used for generating junit report files, which are consumable by jenkins for easier viewing of which tests failed (similar to what Travis does, but without running 3 separate jobs). More info: https://github.com/bats-core/bats-core/issues/107
2. I didn't refactor cluster-sync to be shared between jenkins and travis yet. I might do that as part of switch to OKD (next step after this).


mmazur:
Done. Please disregard travis fails, it's #240 again. This PR only touches Jenkins.

PR #261, opened 2019-06-25
[WIP] Jenkins: add okd4 target

mmazur:
This isn't really WIP, it's only marked that because it's used for debugging #263 and unmergable before that issue gets fixed.

BEFORE MERGE: drop the commit marked TEMPORARY


pkliczewski:
@mmazur Do we want to finalize this PR or is it OK to close?

mmazur:
Closing. Nowadays these tests are regularly run against openshift4 as part of a different pipeline, so this isn't a priority for us.

PR #262, opened 2019-06-26
Workaround for travis+minikube instability

mmazur:
What this does: sleep for 90 seconds before running a new playbook if env == travis. This gives minikube time to recover after whatever's wrong with it and then correctly run the next playbook.

### DIFF
--- a/tests/playbooks/known_issues.yaml
+++ b/tests/playbooks/known_issues.yaml
@@ -18,3 +18,8 @@
   meta: end_play
   when: "skip_{{ current_playbook }} == true"
 
+# This is a workaround for https://github.com/kubevirt/ansible-kubevirt-modules/issues/240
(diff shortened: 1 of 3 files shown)


tests/playbooks/known_issues.yaml
```diff
@@ -18,3 +18,8 @@
   meta: end_play
   when: "skip_{{ current_playbook }} == true"
 
+# This is a workaround for https://github.com/kubevirt/ansible-kubevirt-modules/issues/240
+- name: "Travis/minikube: give cluster time to recover from its instability"
```
machacekondra: Why is this only in known_issues?

mmazur: I'll file an upstream (minikube) bug for this, but need to take care of other stuff first.

Issue #263, opened 2019-06-27
OKD4: api server produces output unparsable by urllib3


mmazur:
```
TASK [Create a blank PVC image] *************************************************************************************************************************************************************
task path: /home/mmazur/dev/ansible-kubevirt-modules/tests/playbooks/kubevirt_pvc.yml:19
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: mmazur
<127.0.0.1> EXEC /bin/sh -c 'echo ~mmazur && sleep 0'
<127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /home/mmazur/.ansible/tmp/ansible-tmp-1561643974.248798-270436959413042 `" && echo ansible-tmp-1561643974.248798-270436959413042="` echo /home/mmazur/.ansible/tmp/ansible-tmp-1561643974.248798-270436959413042 `" ) && sleep 0'
Using module file /home/mmazur/.local/lib/python3.7/site-packages/ansible/modules/cloud/kubevirt/kubevirt_pvc.py
<127.0.0.1> PUT /home/mmazur/.ansible/tmp/ansible-local-21229f2od_f7g/tmpvnbcc3rd TO /home/mmazur/.ansible/tmp/ansible-tmp-1561643974.248798-270436959413042/AnsiballZ_kubevirt_pvc.py
<127.0.0.1> EXEC /bin/sh -c 'chmod u+x /home/mmazur/.ansible/tmp/ansible-tmp-1561643974.248798-270436959413042/ /home/mmazur/.ansible/tmp/ansible-tmp-1561643974.248798-270436959413042/AnsiballZ_kubevirt_pvc.py && sleep 0'
<127.0.0.1> EXEC /bin/sh -c '/usr/bin/python3 /home/mmazur/.ansible/tmp/ansible-tmp-1561643974.248798-270436959413042/AnsiballZ_kubevirt_pvc.py && sleep 0'
<127.0.0.1> EXEC /bin/sh -c 'rm -f -r /home/mmazur/.ansible/tmp/ansible-tmp-1561643974.248798-270436959413042/ > /dev/null 2>&1 && sleep 0'
The full traceback is:
Traceback (most recent call last):
  File "/home/mmazur/.local/lib/python3.7/site-packages/urllib3/response.py", line 603, in _update_chunk_length
    self.chunk_left = int(line, 16)
ValueError: invalid literal for int() with base 16: b''

[...]

PR #264, opened 2019-06-27
Jenkins: switch to openshift 3

mmazur:
Our modules are too buggy with OKD4 for now. Doesn't mean we can't start testing the template modules.


pkliczewski:
Is this because of our modules are buggy or OKD4 being not stable? @mmazur If there are any bugs in the modules please open issues.

mmazur:
Ok, please merge. We're not getting OKD4 yet (because #263), but good old openshift 3 is enough to get us automatic template module testing.

PR #267, opened 2019-06-28
Add CI testing of vm_image role

mmazur:
I skipped it when adding the role PR. Now it's back!


mmazur:
All checks passed, patch is size/S and quick to review, plz mrg.

tests/roles/vm_image.yml
```diff
@@ -0,0 +1,22 @@
...
+        namespace: default
+        size: 100Mi
+        source: blank
+
+      - name: volume3
+        namespace: default
+        size: 100Mi
+
+  roles:
+    - vm_image
```
machacekondra: Maybe add also some cleanup playbook? So we don't waste with resources?

mmazur: The role tests are run only on Travis in a dedicated job and the dedicated VM gets immediately removed afterwards. If we ever get to some more elaborate role testing – sure, might be needed to add a removal step (and maybe stricter verification steps, more in line with what, say, `kubevirt_vm.yml` has). But for now adding object removal would only make the tests run a tiny bit longer without any gain.

PR #272, opened 2019-07-17
Drop pvc cloning and obsolete known issues

mmazur:
Regarding pvc cloning – this used to be possible with CDI, but with 1.9.4 cloning got limited to datavols only and won't be reintroduced for PVCs. However in k8s 1.15 spec.dataSource got introduced and it does the same thing. So once kubevirt_pvc gets support for that, there should again be a test for it here.

### DIFF
--- a/tests/playbooks/known_issues.yaml
+++ b/tests/playbooks/known_issues.yaml
@@ -17,18 +17,6 @@
     skip_template: false
   when: lookup('env','KUBEVIRT_PROVIDER') is match('okd-.*') or lookup('env','KUBEVIRT_PROVIDER') is match('os-.*')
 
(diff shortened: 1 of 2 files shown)


mmazur:
Please merge. This fixes the CI failures on all targets except for branch/devel, as that has a [different issue](https://bugzilla.redhat.com/show_bug.cgi?id=1730706) that I'll address in a separate PR.

pkliczewski:
@mmazur The PR looks good but CI seems to be failing. Please check

mmazur:
@pkliczewski what still fails is a separate issue, which Vatsal created a bugzilla case for already. There's info in the previous comment. I'll fix it in a separate PR.

PR #274, opened 2019-07-29
Stop CI complaining about latest CDI

mmazur:
Force previous CDI version for the time being. I'll add support for latest version in a separate patch.


pkliczewski:
let's wait on CI

mmazur:
Oh, great, seems cdi 1.10 is also not backward compatible in some ways…

mmazur:
@machacekondra please merge, so we stop getting the failmails. I'll figure out what to do about supporting cdi 1.10 separately.

PR #275, opened 2019-08-06
Make CI green again

mmazur:
1. Some updates for kubevirt 0.20.x
2. When using cirros images, use the one tagged with the $KUBEVIRT_VER we're using instead of :latest, cause if there's an incompatibility, :latest will not work (like it doesn't now due to https://github.com/kubevirt/kubevirt/pull/2395)
3. Make it easy to use older kubevirt/cdi versions if latest is borked.


mmazur:
Trying to trigger travis.

mmazur:
Riiight. So this fixes a bunch of issues, but, as it turns out, ansible branch devel introduces some new regression, which still makes it not all green. But at least more green.

So this can either be merged as–is and I'll deal with the ansible/devel issue after my PTO in a separate PR or this PR can just wait for a bit.

mmazur:
@pkliczewski @machacekondra please merge this. It's not green, but only for devel branch and to verify [my fix for devel branch](https://github.com/ansible/ansible/pull/61020) actually works, I need this merged first.

pkliczewski:
> @pkliczewski @machacekondra please merge this. It's not green, but only for devel branch and to verify [my fix for devel branch](https://github.com/ansible/ansible/pull/61020) actually works, I need this merged first.

@mmazur Let's wait for ci job for you fix

mmazur:
@pkliczewski a part of verifying https://github.com/ansible/ansible/pull/61020 is actually correct is running the travis jobs from this repo against https://github.com/ansible/ansible/pull/61020. But for that to actually have a chance of returning green, this PR needs to be merged first, as everything is always red otherwise.

The CI job on https://github.com/ansible/ansible/pull/61020 has no relation to this PR.

